PR #360, opened 2026-04-17
Fix batch transcript metadata and preserve rerun assets

### LINKED ISSUE #359: Fix batch transcript metadata drift and preserve rerun assets
## Problem

Batch retranscription currently has two concrete problems:

1. `session.json` is not refreshed from the saved final transcript, which leaves the Notes sidebar showing stale meeting times and `0 utterances` even when `transcript.final.jsonl` exists.
2. The batch path ignores the effective system-audio sample rate that the recorder already computes to correct process-tap drift. That makes the final mic/system interleave more prone to timing drift.
3. Batch stems and `batch-meta.json` are deleted immediately after final transcript generation, which makes true reruns/debugging impossible for completed sessions.

## Proposed fix

- refresh `session.json` when `transcript.final.jsonl` is written
- carry `sysEffectiveSampleRate` from `AudioRecorder` into batch metadata and use it during batch timestamp reconstruction
- retain batch stems/metadata for a bounded rerun window instead of deleting them immediately

## Validation

- targeted `SessionRepositoryTests` for metadata refresh and retention behavior
- targeted `AudioRecorderTests` for effective sample rate tracking

kkarimi:
Fixes #359

## Summary

- refresh `session.json` from `transcript.final.jsonl` so the saved session metadata matches the batch transcript
- carry the recorder's effective system sample rate into batch metadata and use it when reconstructing system-audio timestamps
- retain batch stems and `batch-meta.json` for a bounded rerun/debug window instead of deleting them immediately after batch transcription

## Why

This fixes two concrete post-meeting issues:

- Notes/Past Meetings can show stale meeting times and `0 utterances` even though a final transcript exists
- completed sessions cannot be truly rerun/debugged because the original batch stems are deleted immediately

The sample-rate change also fixes a real timing bug in the batch path by reusing the recorder's drift correction during final transcript reconstruction.

## Validation

- `swift test --package-path OpenOats --filter AudioRecorderTests`

### DIFF
--- a/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
+++ b/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
@@ -448,7 +448,8 @@ final class LiveSessionController {
                         micStartDate: anchorsData.micStartDate,
                         sysStartDate: anchorsData.sysStartDate,
                         micAnchors: anchorsData.micAnchors,
-                        sysAnchors: anchorsData.sysAnchors
+                        sysAnchors: anchorsData.sysAnchors,
+                        sysEffectiveSampleRate: anchorsData.sysEffectiveSampleRate
                     )
                 )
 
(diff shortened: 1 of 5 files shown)


yazins-ai:
Thanks for another quality contribution, @kkarimi! The metadata refresh after batch transcription and the effective-sample-rate propagation are both solid fixes — and the 7-day retention window for batch stems is a nice improvement for debugging. Tests cover the new behavior well. Merged.

PR #362, opened 2026-04-17
Hide internal Granola tags and group Notes imports by source

### LINKED ISSUE #361: Hide internal Granola import tags and group Notes by source
## Problem

Granola imports currently leak their internal sync tag (`granola:<note-id>`) into the Notes UI:

- the top tag filter bar shows raw Granola note ids
- imported sessions show those raw ids as chips in the sidebar
- imported sessions are mixed into the main list without a clear native grouping

This makes the Notes window feel like it is exposing internal import metadata rather than user-facing organization.

## Proposed fix

- treat `granola:` tags as internal import metadata, not user-visible tags
- preserve those internal tags when a user edits visible session tags
- group the Notes sidebar by source so imported Granola sessions appear together in a native sectioned list

## Validation

- `NotesControllerTests` for hidden internal tags and source grouping
- `SessionRepositoryTests` for preserving hidden Granola linkage when tags are edited

kkarimi:
Fixes #361

## Summary

- hide internal `granola:<note-id>` sync tags from the Notes tag bar, tag chips, and tag autocomplete
- preserve those hidden Granola tags when a user edits visible session tags
- group the Notes sidebar by session source so imported Granola meetings appear together in a native sectioned list

## Why

Granola import ids are implementation detail, not user-facing organization. Exposing them in the Notes UI makes the app feel like it is leaking raw import metadata, and imported sessions are harder to scan because they are not grouped together.

## Validation

### DIFF
--- a/OpenOats/Sources/OpenOats/App/NotesController.swift
+++ b/OpenOats/Sources/OpenOats/App/NotesController.swift
@@ -43,6 +43,12 @@ enum GenerationStatus: Equatable {
     case error(String)
 }
 
+struct SessionSourceGroup: Identifiable {
+    let id: String
+    let title: String
+    let sessions: [SessionIndex]
+}
+
 // MARK: - Controller
 
 /// Owns all notes/history business logic previously embedded in NotesView.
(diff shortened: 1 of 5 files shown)


yazins-ai (approved):
Reviewed the diff, tests, and overall approach. This is clean UX work — hiding internal `granola:` tags while preserving them in storage is the right call, and the source-grouped sidebar is a natural complement.

**Code quality**: Good separation — filtering logic lives in `NotesController`, storage preservation in `SessionRepository`, and the view layer adapts accordingly. The `isUserVisibleSessionTag` / `isInternalSessionTag` pair keeps the tag-hiding rule in one place per layer.

**Tag preservation**: The `updateSessionTags` change correctly splits internal from visible tags and recombines them, so user edits never silently drop the Granola linkage. The test `testUpdateSessionTagsPreservesInternalGranolaTag` confirms this path.

**Source grouping**: `showsSourceSections` correctly avoids adding sections when all sessions are native OpenOats — no visual change for users who don't import. The sort order (OpenOats first, Granola second, others alphabetical) is sensible.

**One note on view duplication**: The context menu + popover code in `NotesView.swift` is duplicated between the sectioned and non-sectioned branches. This is a common SwiftUI trade-off and doesn't block the PR, but if you revisit this area you could extract the session row + context menu into a helper to reduce the repetition.

**Blast radius**: 3 production files + 2 test files, all within the Notes/session management area. No new settings, permissions, or dependencies. No off-device data flow changes.

Waiting on `validate-swift` to go green, then this can merge.

PR #391, opened 2026-04-19
Link renamed meetings into history families

### LINKED ISSUE #390: Link renamed meetings into the same history family
## Summary
Renamed recurring meetings currently lose their history because meeting families are matched by normalized title equality only.

## Proposed fix
- suggest likely older meeting families when an upcoming meeting has no exact history
- let the user explicitly link an older title into the current meeting family
- persist that alias so future history and prep notes resolve through the same family key

## Why now
This keeps the new meeting-history flow useful when calendar titles evolve over time without introducing broad fuzzy matching.

kkarimi:
Fixes #390

## Summary
- add settings-backed meeting history aliases so renamed recurring meetings can resolve into the same family
- suggest likely older meeting families when an upcoming meeting has no exact history and let the user link them explicitly
- canonicalize meeting prep notes through the alias map so prep state follows the linked family

## Validation
- `swift test --filter SettingsStoreTests --filter UpcomingCalendarGroupingTests --filter NotesControllerTests`

## Screenshot
Meeting history alias linking

### DIFF
--- a/OpenOats/Sources/OpenOats/App/NotesController.swift
+++ b/OpenOats/Sources/OpenOats/App/NotesController.swift
@@ -10,6 +10,8 @@ struct NotesState {
     var selectedSessionID: String?
     var selectedMeetingHistory: MeetingHistorySelection?
     var meetingHistoryEntries: [MeetingHistoryEntry] = []
+    var relatedMeetingSuggestions: [MeetingHistorySuggestion] = []
+    var linkingMeetingSuggestionKey: String?
     var loadedTranscript: [SessionRecord] = []
     var loadedNotes: GeneratedNotes?
     var loadedCalendarEvent: CalendarEvent?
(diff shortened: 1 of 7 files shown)


yazins-ai (approved):
Clean implementation. The alias chain resolution with cycle detection, token-based suggestion scoring, and the dominant-family filter for single-token queries are all well thought through. Tests cover the key scenarios.

A few notes from the review:

- **Alias persistence**: JSON-encoded `[String: String]` in UserDefaults matches the existing `meetingPrepNotesByKey` pattern — consistent.
- **Canonical key resolution**: The `seen` set in `canonicalHistoryKey` prevents infinite loops if aliases ever form a cycle — good defensive coding.
- **Suggestion thresholds**: Requiring >= 4 character tokens for single-token matching and >= 75% overlap for multi-token matching are sensible defaults to avoid false positives.
- **Prep notes canonicalization**: Routing `meetingPrepNotes(for:)` and `setMeetingPrepNotes(_:for:)` through canonical keys means prep notes follow the alias automatically — nice.

PR #411, opened 2026-04-20
Preserve calendar metadata when saving final transcripts

### LINKED ISSUE #410: Preserve calendar metadata when saving final transcripts
## Summary
Saving `transcript.final.jsonl` currently refreshes `session.json` and drops the existing `calendarEvent` payload.

## Problem
Sessions that started with calendar metadata lose their calendar/meeting info after final transcript save, so the Notes view can no longer show the meeting context strip even though the session title is still correct.

## Fix
Preserve the existing `calendarEvent` when refreshing session metadata after saving the final transcript, and cover it with a regression test.

kkarimi:
Fixes #410

## Summary
- preserve the existing calendar event when refreshing `session.json` after writing `transcript.final.jsonl`
- add a regression test covering finalize -> save final transcript -> load session

## Validation

### DIFF
--- a/OpenOats/Sources/OpenOats/Storage/SessionRepository.swift
+++ b/OpenOats/Sources/OpenOats/Storage/SessionRepository.swift
@@ -504,7 +504,8 @@ actor SessionRepository {
                 engine: meta.engine,
                 tags: meta.tags,
                 folderPath: meta.folderPath,
-                source: meta.source
+                source: meta.source,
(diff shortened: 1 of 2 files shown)


yazins-ai:
Merged — thanks @kkarimi for the clean fix and the regression test. The missing `calendarEvent` passthrough was a real bug that would silently strip meeting context from finalized sessions.

PR #422, opened 2026-04-21
Fix local model cache path stability across app updates

### LINKED ISSUE #376: Prevent redownloading model after each version update
A bit annoying to download model after each version update. Is there a way to keep it persistent somewhere?

kkarimi:
## Summary
- move Parakeet and Qwen local model resolution onto an OpenOats-owned stable cache root
- migrate existing FluidAudio cache directories into the stable OpenOats path before checking readiness or loading
- clear both stable and pre-migration cache locations when the user explicitly clears a local model cache
- add focused migration tests for the stable model store

## Why
Users can end up being prompted to re-download local transcription models after upgrading because the app inherits FluidAudio cache folder naming changes. This keeps future model paths stable from the app side and performs a one-time migration into those paths.

Closes #376

## Validation
- swift test --package-path OpenOats --filter OpenOatsLocalModelStoreTests
- swift test --package-path OpenOats --filter TranscriptionBackendTests
- swift test --package-path OpenOats --filter TranscriptionEngineTests

### DIFF
--- a/OpenOats/Sources/OpenOats/Transcription/OpenOatsLocalModelStore.swift
+++ b/OpenOats/Sources/OpenOats/Transcription/OpenOatsLocalModelStore.swift
@@ -1,0 +1,253 @@
+import FluidAudio
+import Foundation
+
+enum OpenOatsLocalModelStore {
+    static func parakeetDirectory(
+        for version: AsrModelVersion,
+        baseDirectory: URL = defaultBaseDirectory(),
+        fluidAudioModelsRoot: URL = MLModelConfigurationUtils.defaultModelsDirectory()
+    ) -> URL {
+        let wrapper = baseDirectory
+            .appendingPathComponent("parakeet", isDirectory: true)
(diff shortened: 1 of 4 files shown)


yazins-ai (approved):
Clean implementation. The new `OpenOatsLocalModelStore` gives OpenOats stable, app-owned cache paths for Parakeet and Qwen3 models, decoupling them from FluidAudio's internal directory naming. The lazy migration on first access is the right approach — existing model files move transparently without re-download.

Code highlights:
- `moveDirectoryIfNeeded` correctly handles the move→fallback-to-copy path and guards against same-source/dest
- `clearParakeetCache` / `clearQwen3Cache` clean both stable and legacy locations — good for a complete clear
- Migration candidate deduplication via `Set<String>` avoids double-processing when current and legacy paths resolve identically
- Tests cover the core migration, download-into-stable, and clear-both-locations paths

All CI checks pass (validate-swift, package-smoke, ui-smoke).

PR #425, opened 2026-04-21
Add default folders for meeting families

### LINKED ISSUE #424: Add default folders for meeting families
## Summary
Users can already set template defaults per meeting family. Folder defaults should work the same way so recurring meetings land in the right place without manual filing.

## What to add
- Persist a default folder path per meeting family
- Show the folder control in the meeting-family header metadata row
- Use folder color in the control when a default is set
- Allow creating/selecting folders up to one subfolder deep
- When changing an existing family default, ask whether to affect only future meetings or move saved meetings too
- Apply the saved folder automatically when future sessions in that family finalize

## Notes
- Keep the two-level folder limit for this UI: top-level folder plus optional one subfolder
- Do not widen this into a general folder-management screen in the same change

kkarimi:
Closes #424

## Summary
- add a meeting-family default folder preference in settings
- surface the folder picker in the meeting-family header metadata row with folder color
- confirm whether a folder change should apply only to future meetings or also move saved meetings
- auto-apply the saved folder when future sessions in that family finalize

## UI
- keep the folder control in the header metadata row near time and calendar
- remove the visible `Prep notes` label and `Calendar:` prefixes from the same card
- keep the two-level folder limit in this workflow (`Work` or `Work/1:1s`)

## Validation
- `swift test --package-path OpenOats --filter SettingsStoreTests`

### DIFF
--- a/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
+++ b/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
@@ -404,6 +404,12 @@ final class LiveSessionController {
             )
         )
 
+        if let settings,
+           let event = endingMetadata?.calendarEvent,
+           let folderPath = settings.meetingFamilyPreferences(for: event)?.folderPath {
+            await coordinator.sessionRepository.updateSessionFolder(sessionID: sessionID, folderPath: folderPath)
+        }
+
         // 5. Build index for UI state
         let index = SessionIndex(
             id: sessionID,
--- a/OpenOats/Sources/OpenOats/App/NotesController.swift
+++ b/OpenOats/Sources/OpenOats/App/NotesController.swift
@@ -801,6 +801,39 @@ final class NotesController {
         }
     }
 
+    func setMeetingFamilyFolderPreference(_ folderPath: String?) {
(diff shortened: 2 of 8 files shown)


yazins-ai (approved):
Well-structured feature that extends the existing meeting-family preferences to include folder defaults.

Code highlights:
- Reuses `NotesFolderDefinition.normalizePath()` for validation, with an additional two-level depth guard in both `SettingsStore` and the view layer
- Confirmation dialog when changing folders for families with existing sessions is good UX — "future only" vs "move saved meetings too"
- `applyMeetingFamilyFolderPreference` in NotesController correctly resolves through meeting history aliases before persisting
- LiveSessionController auto-applies the folder default when a session finalizes, which is the right integration point
- Prep notes label replaced with placeholder text is a nice polish — reduces visual clutter while keeping discoverability
- Tests cover alias canonicalization, path depth rejection, and the move-existing-sessions flow

The folder editor sheet was cleanly extracted into a shared `folderEditorSheet` builder, avoiding duplication between the session and meeting-family use cases.

All CI checks pass (validate-swift, package-smoke, ui-smoke).

PR #428, opened 2026-04-21
Show knowledge base coverage for meeting families

### LINKED ISSUE #427: Surface knowledge base coverage for meeting families
## Summary
Meeting families are now a stable context layer with history, folder defaults, and template defaults. The next useful step is to show whether OpenOats already has relevant knowledge-base material for a meeting family.

## Smallest useful slice
- run a lightweight title-only KB retrieval pass when a meeting family is shown
- surface a compact availability signal in the meeting-family header
- keep the signal passive for now; do not yet pin documents or feed them into note generation automatically

## Notes
- this should be honest UI, not an implied guarantee that the matched docs are already being used automatically
- document counts should be de-emphasized in the visible label; the point is availability, not scoring

kkarimi:
Closes #427

## Summary
- run a lightweight title-only knowledge-base lookup when a meeting family is shown
- surface a compact `Knowledge available` signal in the meeting-family header
- keep the signal visible when focusing a past session inside the same family
- dedupe matches by document so the signal reflects coverage rather than chunk count noise

## Notes
- this is a passive coverage signal only
- it does not yet pin documents or inject them automatically into meeting-family notes generation
- the label intentionally emphasizes availability rather than document counts

## Validation

### DIFF
--- a/OpenOats/Sources/OpenOats/App/NotesController.swift
+++ b/OpenOats/Sources/OpenOats/App/NotesController.swift
@@ -10,6 +10,8 @@ struct NotesState {
     var selectedSessionID: String?
     var selectedMeetingFamily: MeetingFamilySelection?
     var meetingHistoryEntries: [MeetingHistoryEntry] = []
+    var meetingFamilyKnowledgeBaseCoverage: MeetingFamilyKnowledgeBaseCoverage?
+    var isMeetingFamilyKnowledgeBaseLoading: Bool = false
     var relatedMeetingSuggestions: [MeetingHistorySuggestion] = []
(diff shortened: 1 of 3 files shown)


yazins-ai (approved):
Clean implementation. The new KB coverage signal follows existing patterns well:

- Correctly uses `searchContextPacks` with proper guard clauses (KB configured, indexed, non-empty title)
- Task management and cancellation mirrors the existing `meetingHistoryPreviewTask` pattern
- Deduplication by document path keeps the signal honest — coverage, not chunk noise
- Passive badge with tooltip is the right UX for a first pass
- Tests cover the static dedup logic

All CI checks green. Merging.

PR #434, opened 2026-04-21
fix: raise LLM request timeout to 300s to unblock slow completions

BJonny:
## Summary

- `OpenRouterClient` was constructing `URLRequest` without setting `timeoutInterval`, so it silently inherited Foundation's default of 60 seconds. For `URLSession.bytes(for:)` that default acts as the **idle-between-bytes** timeout, not a total budget — any >60s gap before the next chunk kills an otherwise healthy stream.
- Notes generation hits this reliably against cold Ollama/MLX models, reasoning models with long hidden chain-of-thought, and long transcripts where first-token latency exceeds 60s. Symptom: `URLError.timedOut` surfacing as `notesGenerationStatus = .error(...)` in `NotesView`.
- Fix sets `urlRequest.timeoutInterval = 300` on both the streaming path (`streamCompletion`) and the non-streaming path (`complete`, used by the gate / judge / structured-JSON callers in the suggestion + sidecast pipelines), which are vulnerable to the same default.

## Why 300s

- Long enough to cover realistic cold-start + first-token latency on local Ollama/MLX and reasoning models with extended thinking budgets.
- Short enough to give up on a genuinely dead TCP connection.
- For streaming, remember this is the *idle* timeout between bytes — the clock resets on each chunk, so it does not cap legitimately long completions.

## Test plan

- [x] `swift build -c debug` — clean
- [x] `./scripts/build_swift_app.sh` — builds, signs, installs to `/Applications/OpenOats.app`
- [ ] Reproduce original bug: point notes generation at a freshly-loaded large Ollama model (e.g. `llama3.1:70b`) or an OpenRouter reasoning model on a long transcript and confirm 60s timeout no longer fires
- [ ] Confirm no regressions on a fast cloud model (suggestions / sidecast gate still responsive)
- [ ] Optional: add a unit test that injects a stream which pauses >60s before yielding bytes and asserts the stream completes

### DIFF
--- a/OpenOats/Sources/OpenOats/Intelligence/OpenRouterClient.swift
+++ b/OpenOats/Sources/OpenOats/Intelligence/OpenRouterClient.swift
@@ -77,6 +77,10 @@ actor OpenRouterClient {
 
                     var urlRequest = URLRequest(url: targetURL)
                     urlRequest.httpMethod = "POST"
+                    // Idle timeout between streamed bytes. Must cover cold-start of local models
+                    // (Ollama/MLX) and first-token latency of reasoning models, which routinely
+                    // exceed URLRequest's 60s default.
+                    urlRequest.timeoutInterval = 300
                     urlRequest.setValue("application/json", forHTTPHeaderField: "Content-Type")
                     if let apiKey, !apiKey.isEmpty {
                         urlRequest.setValue("Bearer \(apiKey)", forHTTPHeaderField: "Authorization")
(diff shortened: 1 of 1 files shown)


yazins-ai (approved):
Good catch. `OpenRouterClient` was the only LLM client without an explicit timeout — the rest of the codebase already sets `timeoutInterval` where needed (e.g. `OllamaSetupClient` at 3600s, `ElevenLabsScribeBackend` at 30s, `APIKeyValidator` at 5s).

The fix is minimal and correct:

- **Streaming path** (`streamCompletion`): 300s acts as the idle-between-chunks timeout, which is the right semantic for `URLSession.bytes(for:)`. Cold-start latency on local models easily exceeds 60s.
- **Non-streaming path** (`complete`): 300s covers gate/judge/structured-JSON calls hitting slow backends.

One file, no new dependencies, no new settings, no behavior change beyond the timeout value. Ready to merge once CI is green.

yazins-ai:
Merged — thanks @BJonny for the clean fix! Raising the idle timeout to 300s is the right call for cold local models and reasoning models with extended thinking. Appreciate the thorough write-up covering both the streaming and non-streaming paths.

PR #435, opened 2026-04-22
Fix transcription settings tab crash when enumerating devices

### LINKED ISSUE #337: Crash in settings - transcription tab
Hi! 
v 1.47.0 (and 1.46 I guess) crashes often when I click on transcription tab. 

Steps to reproduce:

[crash.log](https://github.com/user-attachments/files/26630776/crash.log)

 enter settings, click "transcription" tab

macOS: 26.4 (25E246)
Hardware: Mac mini M4, 16 GB RAM (Mac16,10)

The crash happens inside `TranscriptionSettingsTab.body` during view construction. The signature (EXC_BREAKPOINT in the Swift allocator) points to heap corruption — most likely something being written off the main actor while the view reads it.
Full crash log attached

kkarimi:
Fixes #337.

## Summary
- correct CoreAudio CFString ownership in microphone device enumeration
- correct CoreAudio CFString ownership in output device enumeration
- avoid over-releasing device name and UID values while building the transcription settings tab

## Why
Opening the Transcription settings tab enumerates input/output devices on appear. Those code paths were using `takeRetainedValue()` for CoreAudio-provided CFStrings, which is the wrong ownership assumption here and is a plausible source of the malloc corruption reported in #337.

## Validation
- `swift build -c debug`
- `swift test` *(current main still has unrelated `MeetingDetectorTests` failures)*

### DIFF
--- a/OpenOats/Sources/OpenOats/Audio/MicCapture.swift
+++ b/OpenOats/Sources/OpenOats/Audio/MicCapture.swift
@@ -319,7 +319,7 @@ final class MicCapture: @unchecked Sendable {
             status = AudioObjectGetPropertyData(deviceID, &nameAddress, 0, nil, &nameSize, &name)
             guard status == noErr, let name else { continue }
 
-            result.append((id: deviceID, name: name.takeRetainedValue() as String))
+            result.append((id: deviceID, name: name.takeUnretainedValue() as String))
         }
 
         return result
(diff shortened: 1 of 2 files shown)


yazins-ai (approved):
Correct fix. CoreAudio's `AudioObjectGetPropertyData` follows the Get Rule — the caller does not own the returned CFString, so `takeUnretainedValue()` is the right call. The previous `takeRetainedValue()` over-released the string, which matches the heap corruption crash in #337.

Changes are minimal (4 single-word substitutions across 2 files in the same subsystem), consistent, and well-explained.

Waiting on `validate-swift` to go green before merge.

PR #438, opened 2026-04-22
Make knowledge base indexing lazy on startup

### LINKED ISSUE #437: Avoid eager knowledge base reindexing on startup
## Problem
If a Knowledge Base folder is configured, OpenOats currently clears and reindexes it during app startup. That makes launch heavier than it needs to be, can trigger unnecessary provider work, and front-loads Keychain/API key access even before the user enters a KB-backed flow.

## Proposed fix
- restore KB state from the on-disk cache on startup when the cache still matches the configured folder and embedding settings
- lazy-load or reindex only when KB-backed retrieval is actually used
- still force a real reindex when the KB folder or embedding configuration changes

## Notes
Meeting-family KB coverage should participate in the same lazy path instead of requiring the KB to already be hot.

kkarimi:
Closes #437

## Summary
- stop forcing a full knowledge-base reindex during app startup
- restore cached KB state on first sync when the cache still matches the configured folder and embedding settings
- let KB-backed retrieval paths lazily load from cache or index on demand
- keep real reindexing for actual KB config changes such as folder, provider, or loaded Voyage key changes

## Why
The previous startup path treated a configured KB folder as a signal to immediately clear and rebuild the KB. That was too eager for the current product shape and could front-load provider work and secret access before the user entered any KB-backed flow.

## Validation
- `swift test --package-path OpenOats --filter 'LiveSessionControllerTests|KnowledgeBaseTests|NotesControllerTests'`\n- `SKIP_SIGN=1 SKIP_INSTALL=1 ./scripts/build_swift_app.sh`

### DIFF
--- a/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
+++ b/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
@@ -56,9 +56,10 @@ final class LiveSessionController {
     private var observedAudioLevel: Float = 0
     private var observedSuggestions: [Suggestion] = []
     private var observedIsGenerating = false
-    private var observedKBFolderPath = ""
+    private var observedKBFolderPath: String?
     private var observedNotesFolderPath = ""
-    private var observedVoyageApiKey = ""
+    private var observedEmbeddingProvider: EmbeddingProvider?
+    private var observedVoyageApiKey: String?
     private var observedTranscriptionModel: TranscriptionModel = .parakeetV2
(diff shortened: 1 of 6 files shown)


yazins-ai:
Thanks @kkarimi — this is a well-structured improvement. Deferring KB indexing to first retrieval and restoring from cache on startup is exactly the right approach. The test coverage for the lazy Voyage key behavior is a nice touch too. Merged!

PR #441, opened 2026-04-22
Make startup service initialization lazy

### LINKED ISSUE #440: Lazy-load startup service initialization
## Summary
OpenOats still eagerly initializes the full service graph during app launch, even before the user opens notes, starts a meeting, or imports audio. That does unnecessary startup work and makes the app pay for recording and KB services before they are actually needed.

## Proposed change
- split startup service initialization into view and recording buckets
- remove the eager container initialization from the main window startup task
- initialize meeting services on demand when starting a meeting
- initialize recording services on demand when importing audio
- initialize view services on demand when opening notes

## Why this matters
This keeps startup lighter without changing the actual meeting, notes, or import flows.

kkarimi:
## Summary
- split service initialization into view and recording buckets
- stop eagerly building the full service graph during app startup
- initialize services on demand when starting meetings, importing audio, opening notes, or using the menu bar start action

## Testing
- swift test --package-path OpenOats *(fails on current main in pre-existing `MeetingDetectorTests` only)*

Closes #440

### DIFF
--- a/OpenOats/Sources/OpenOats/App/AppContainer.swift
+++ b/OpenOats/Sources/OpenOats/App/AppContainer.swift
@@ -24,7 +24,8 @@ final class AppContainer {
     private(set) var calendarManager: CalendarManager?
 
     private var didSeedInitialData = false
-    private var didInitializeServices = false
+    private var didInitializeViewServices = false
+    private var didInitializeRecordingServices = false
 
     init(
         mode: AppRuntimeMode,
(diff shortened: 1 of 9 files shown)


yazins-ai (approved):
This extends the lazy initialization pattern from #438 to the full service graph. The split into view and recording service buckets is clean, and the on-demand initialization points are well-chosen:

- `ensureMeetingServicesInitialized` for session start and deep links
- `ensureRecordingServicesInitialized` for audio import and model downloads
- `ensureViewServicesInitialized` for notes view

The double-check pattern in `ensureViewServicesInitialized` (checking `coordinator.knowledgeBase` before the flag) correctly handles the case where the KB was already initialized by #438's lazy loading.

The `onToggleMeeting` callback through `MenuBarController` → `MenuBarPopoverView` is the right approach — it ensures services are initialized via `AppDelegate.toggleMeeting` before the coordinator handles the start/stop event.

Test updates look good — the renamed `testDeepLinkStartInitializesServicesOnDemand` correctly reflects the new behavior where deep links trigger on-demand initialization rather than silently failing.

All CI checks green. Merging.

yazins-ai:
Merged — thanks @kkarimi for the clean refactor! Splitting the service graph into lazy buckets is a solid improvement for startup performance.

PR #442, opened 2026-04-22
fix: send ElevenLabs Scribe keyterms as repeated multipart fields

Newarr:
## What

Fix HTTP 400 on every ElevenLabs Scribe request when Custom Vocabulary is non-empty. Also log response bodies on non-2xx so the next provider-side break isn't opaque.

## Why

The backend I added in #240 encodes `keyterms` as a single multipart field whose value is a JSON-literal array. ElevenLabs expects one part per term ([API ref](https://elevenlabs.io/docs/api-reference/speech-to-text/convert), [JS SDK](https://github.com/elevenlabs/elevenlabs-js)). The server reads our value as one keyterm whose brackets trip the validator:

```json
{"detail":{"code":"invalid_parameters","message":"Some keyword contains invalid characters","param":"keywords"}}
```

Any user with non-empty Custom Vocabulary on Scribe has been silently 400'ing every chunk since #240 shipped. The 400 body was discarded at the error path, so it only showed as `httpError(statusCode: 400)` in logs — hence the second hunk.

## Testing

Multipart construction extracted into `ElevenLabsScribeBackend.buildMultipartBody` (internal static, mirrors the `WAVEncoder` pattern) and covered by two unit tests: one asserts three keyterms produce three separate `name="keyterms"` parts, the other asserts the field is omitted entirely when vocabulary is empty. Full suite: 505 passed, 0 failed on `swift test`.

Live API (2026-04-22, 1.5s PCM16 WAV):

| Request | Result |
|---|---|
| `keyterms=["a","b","c"]` (before) | 400 |
| `keyterms=a`, `keyterms=b`, `keyterms=c` (after) | 200 |

### DIFF
--- a/OpenOats/Sources/OpenOats/Transcription/ElevenLabsScribeBackend.swift
+++ b/OpenOats/Sources/OpenOats/Transcription/ElevenLabsScribeBackend.swift
@@ -74,35 +74,14 @@ final class ElevenLabsScribeBackend: TranscriptionBackend, @unchecked Sendable {
 
         // 2. Build multipart/form-data body
         let boundary = UUID().uuidString
-        var body = Data()
-
-        body.appendMultipart(boundary: boundary, name: "model_id", value: "scribe_v2")
-
         let languageCode = locale.language.languageCode?.identifier ?? ""
-        if !languageCode.isEmpty {
-            body.appendMultipart(boundary: boundary, name: "language_code", value: languageCode)
-        }
-
-        body.appendMultipart(
+        let body = Self.buildMultipartBody(
             boundary: boundary,
-            name: "file",
-            filename: "audio.wav",
-            contentType: "audio/wav",
-            data: wavData
+            wavData: wavData,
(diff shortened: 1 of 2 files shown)


yazins-ai (approved):
Clean fix. The multipart encoding change is correct per the ElevenLabs Speech-to-Text API — each keyterm needs its own part rather than a single JSON-array field.

Code notes:

- Error body logging with `privacy: .private` and the 2048-byte cap is appropriate — provides debug visibility without leaking data in regular log captures
- Extracting `buildMultipartBody` as an internal static makes the fix straightforward to test
- Both test cases are well-targeted: verifying per-term parts and the empty-vocabulary path

All CI checks green. Merging.

yazins-ai:
Merged — thanks @Newarr for tracking this down and fixing it. The per-term multipart encoding matches the ElevenLabs API spec, and the error body logging will make future provider-side issues much easier to diagnose.

PR #445, opened 2026-04-22
Fix live control stopping active recordings

### LINKED ISSUE #446: Pressing the Live status control during a recording stops the session
## Summary
While a meeting is actively recording, the green `Live` control in the main window reads like a status indicator, but it currently acts as the stop action.

## Problem
Users can click `Live` expecting to open or focus the live state and instead silently stop the session. The app then returns to the idle `Coming up` screen, which makes it look like the recording failed or disappeared.

## Expected
- `Live` / `Muted` should behave like a passive status indicator while a meeting is running
- stopping should require an explicit control, clearly labeled as `Stop`

## Actual
- the `Live` pill is the stop button
- clicking it mid-recording ends the session and drops the app back to idle

## Notes
This is a control affordance bug rather than a recording engine failure.

kkarimi:
Fixes #446

## Summary
- make the Live/Muted indicator in the control bar a non-interactive status chip while a meeting is running
- add an explicit red Stop button so stopping an active meeting is intentional
- update the UI smoke test to use the keyboard shortcut and verify the new stop control

## Validation
- ./scripts/build_swift_app.sh
- ./scripts/run_ui_smoke.sh

### DIFF
--- a/OpenOats/Sources/OpenOats/Views/ControlBar.swift
+++ b/OpenOats/Sources/OpenOats/Views/ControlBar.swift
@@ -70,20 +70,35 @@ struct ControlBar: View {
             }
 
             HStack(spacing: 10) {
-                Button(action: onToggle) {
+                if isRunning {
                     HStack(spacing: 6) {
-                        if isRunning {
-                            // Pulsing dot when live, static when muted
-                            Circle()
-                                .fill(isMicMuted ? Color.red : Color.green)
-                                .frame(width: 8, height: 8)
-                                .scaleEffect(isMicMuted ? 1.0 : 1.0 + CGFloat(audioLevel) * 0.5)
-                                .animation(.easeOut(duration: 0.1), value: audioLevel)
+                        // Pulsing dot when live, static when muted
+                        Circle()
+                            .fill(isMicMuted ? Color.red : Color.green)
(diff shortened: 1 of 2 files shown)


yazins-ai (approved):
Clean fix. The separation of the status indicator from the stop action is exactly right — the Live/Muted pill becomes a passive chip while running, and stopping requires an intentional click on an explicit red Stop button.

Code review:

- **ControlBar.swift**: The `isRunning` branch now renders the status chip as a plain `HStack` (no `Button` wrapper), which eliminates the accidental-stop path. The new `Stop` button uses `OpenOatsProminentButtonStyle(color: .red)` for clear visual intent. The `else` branch (idle state) keeps the existing Start capsule behavior unchanged.
- **SmokeTests.swift**: Updated to use the keyboard shortcut (`Cmd+Shift+L`) and verify the new `app.controlBar.stop` identifier, which is a better test since it doesn't rely on the now-removed toggle-as-stop path.
- Accessibility identifiers are well-structured: `status` for the indicator, `stop` for the action, `toggle` for idle start.
- No new dependencies, settings, or permissions.

yazins-ai:
Merged — thanks @kkarimi for separating the status indicator from the stop action. The explicit Stop button is a much safer interaction pattern for active meetings.

PR #448, opened 2026-04-22
Reuse abandoned empty meeting session rows

### LINKED ISSUE #447: Reopening the app during a failed meeting start creates duplicate empty session rows
## Summary
If a meeting start fails or the app is reopened before any transcript is captured, pressing record again creates a brand new session row instead of reusing the abandoned empty row for the same meeting.

## Problem
This leaves duplicate rows for the same meeting family, usually with one empty row and one real row. The common case is:
- press record for a calendar-backed meeting
- the recording never really starts or the app is reopened
- press record again
- OpenOats creates a second session instead of adopting the abandoned empty stub

## Expected
When there is an obviously abandoned session stub for the same meeting family, OpenOats should reuse it instead of creating another row.

## Actual
A new session is always created, because the app does not persist enough meeting identity at session start and has no resume/reuse path for abandoned empty rows.

## Scope
This issue is about safely reusing abandoned empty stubs only. It is not true arbitrary resume of a partially running live pipeline.

kkarimi:
Fixes #447

## Summary
- persist meeting identity on the session row as soon as recording starts
- reuse an abandoned empty session stub for the same meeting family instead of creating a second row
- add regression coverage for repository reuse rules and the live-session start path

## Notes
This is a narrow reuse path for abandoned empty rows only. It does not attempt to resume a partially running live pipeline with real transcript or audio artifacts.

## Validation
- swift test --package-path OpenOats --filter SessionRepositoryTests

### DIFF
--- a/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
+++ b/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
@@ -322,12 +322,18 @@ final class LiveSessionController {
         }
 
         let templateID = coordinator.selectedTemplate?.id
-        let handle = await coordinator.sessionRepository.startSession(
-            config: SessionStartConfig(
-                templateID: templateID,
-                templateSnapshot: coordinator.sessionTemplateSnapshot
-            )
+        let startConfig = SessionStartConfig(
+            templateID: templateID,
+            templateSnapshot: coordinator.sessionTemplateSnapshot,
+            title: metadata.title ?? metadata.calendarEvent?.title,
+            calendarEvent: metadata.calendarEvent
         )
+        let handle: SessionHandle
+        if let resumed = await coordinator.sessionRepository.resumeAbandonedSession(config: startConfig) {
+            handle = resumed
+        } else {
+            handle = await coordinator.sessionRepository.startSession(config: startConfig)
+        }
         _currentSessionID = handle.sessionID
         let initialScratchpad = pendingInitialScratchpad?.trimmingCharacters(in: .whitespacesAndNewlines)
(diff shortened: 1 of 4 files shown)


yazins-ai (approved):
Well-implemented fix with good test coverage. This complements the retroactive ghost-collapse logic from #414 by preventing duplicate rows at session start rather than cleaning them up after the fact.

Code review:

- **SessionStartConfig**: Adding `title` and `calendarEvent` as optional fields with backward-compatible defaults is the right approach — existing callers continue to work unchanged.
- **resumeAbandonedSession**: Correctly resets actor state (`currentSessionID`, `hasReportedWriteError`, `liveUtteranceCount`), opens the live transcript handle, and updates metadata with fresh template/title/calendar data. The `seekToEndOfFile()` in `openLiveTranscriptFileHandle` is correct for the resume case.
- **resumableSessionID**: The matching logic is solid — exact calendar event ID match first, then fuzzy match via `MeetingHistoryResolver.historyKey`. The safety guards (no utterances, no notes, no meaningful artifacts, 6-hour max gap) are conservative and appropriate. Only truly empty abandoned stubs get reused.
- **LiveSessionController**: Clean integration — tries resume before creating a fresh session, using the same `SessionStartConfig` for both paths.
- **Tests**: Good coverage of the happy path (stub reuse), artifact-skip case (sessions with transcript are left alone), identity persistence, and controller-level integration.

yazins-ai:
Merged — thanks @kkarimi for tracking down the abandoned-stub duplication path and adding solid regression coverage for both the repository reuse rules and the live-session controller.

PR #478, opened 2026-04-24
Show earlier meetings from today in Coming up

### LINKED ISSUE #477: Show earlier meetings from today in Coming up
## Problem

The `Coming up` dashboard only shows the current meeting and future meetings. If a user wants to work with a meeting that already happened earlier today, there is no natural entry point from the home screen.

This blocks the next slice of work where users should be able to add their own transcript to a past meeting.

## Proposed fix

Add a lightweight `Earlier today` disclosure inside the `Today` section of `Coming up` so ended meetings from earlier in the day can be revealed without disrupting the normal current/upcoming list.

## Scope

- fetch ended meetings from earlier today
- show them behind a subtle disclosure in the `Today` group
- keep notes/folder affordances on those rows
- hide join for past rows

kkarimi:
Fixes #477

## Summary

- add an `Earlier today` disclosure inside the `Today` section of `Coming up`
- fetch ended meetings from earlier in the same day separately from the normal upcoming selection
- keep notes/folder affordances on those older rows while hiding join
- use a dashed separator to distinguish earlier-today rows from the current/upcoming list

## Why

This is the first slice toward letting users add their own transcript to a past meeting. Before that workflow exists, the dashboard needs a natural way to reveal meetings that already happened earlier today.

### Screenshot
Screenshot 2026-04-24 at 14 37 08

## Validation

- `swift test --package-path OpenOats --filter UpcomingCalendarGroupingTests`

### DIFF
--- a/OpenOats/Sources/OpenOats/Meeting/CalendarManager.swift
+++ b/OpenOats/Sources/OpenOats/Meeting/CalendarManager.swift
@@ -94,6 +94,29 @@ final class CalendarManager {
         return events.map { CalendarEvent(from: $0) }
     }
 
+    /// Calendar events occurring on the same local day as the given date, ordered by start date.
+    /// Returns an empty array if access is not authorized.
+    func events(onSameDayAs date: Date = Date()) -> [CalendarEvent] {
+        guard accessState == .authorized else { return [] }
+        let calendars = eventCalendars()
+        let calendar = Calendar.current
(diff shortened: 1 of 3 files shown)


yazins-ai:
Squash-merged — thanks @kkarimi! The disclosure approach is clean, and the test coverage on the selection logic is appreciated. Including the cask version bump to 1.62.0 in a follow-up commit on main.

PR #481, opened 2026-04-24
Add manual transcript import for no-transcript sessions

### LINKED ISSUE #480: Allow adding manual transcripts to no-transcript sessions
## Summary
Sessions that exist in history but have no transcript are currently a dead end in the Transcript tab.

## Problem
- past sessions can exist with calendar context, notes, and meeting-family history but no recorded utterances
- users have no way to paste or import transcript text afterward
- that blocks transcript cleanup, transcript review, and downstream note generation from that session

## First slice
- add an `Add Transcript` action in the Transcript tab empty state for an existing no-transcript session
- store the raw source as `transcript.manual.txt`
- generate a basic `transcript.final.jsonl` from the pasted text
- reload the session so the normal transcript and notes flows work

## Out of scope
- creating a session directly from an `Earlier today` calendar row with no existing session yet
- editing the manual source after import
- file import beyond paste/type text

kkarimi:
Fixes #480

## Summary
- add an `Add Transcript` action in the Transcript tab empty state for existing no-transcript sessions
- persist the pasted raw text as `transcript.manual.txt`
- convert the pasted text into a basic `transcript.final.jsonl` and reload the selected session
- keep the empty-state action quieter with a bordered control instead of a bright primary CTA

## Notes
- speaker parsing supports simple prefixes like `You:`, `Them:`, and `Speaker 2:`
- leading timestamps are stripped during import
- this first slice is for existing session rows only, not direct import from an `Earlier today` calendar row

## Validation
- `./scripts/run_ui_smoke.sh`

### DIFF
--- a/OpenOats/Sources/OpenOats/App/NotesController.swift
+++ b/OpenOats/Sources/OpenOats/App/NotesController.swift
@@ -627,6 +627,25 @@ final class NotesController {
         }
     }
 
+    func addManualTranscript(_ rawText: String) {
+        guard let sessionID = state.selectedSessionID else { return }
+
+        let trimmed = rawText.trimmingCharacters(in: .whitespacesAndNewlines)
+        guard !trimmed.isEmpty else { return }
+
+        let baseDate = state.sessionHistory.first(where: { $0.id == sessionID })?.startedAt
+            ?? state.loadedCalendarEvent?.startDate
+            ?? Date()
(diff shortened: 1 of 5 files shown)


yazins-ai:
Thanks @kkarimi — clean implementation with solid test coverage. The manual transcript import fills a real gap for sessions that have calendar context but no recording. Merged!

PR #484, opened 2026-04-24
Add transcript flow for earlier meetings from today

### LINKED ISSUE #483: Allow adding transcripts for earlier meetings from today
## Problem

Earlier meetings from today disappear from the Coming up dashboard once there are no more remaining meetings later today.

That means there is no path from the dashboard into the manual transcript flow for a meeting that already ended today.

## Expected

- Show an Earlier today section even when today has no remaining upcoming meetings.
- Let an ended meeting from that section open or create its session and route into the Transcript tab.
- If the meeting has no transcript yet, open the Add Transcript flow immediately.

kkarimi:
Fixes #483

## What changed

- add a synthetic Today section when there are earlier meetings from today but no remaining meetings later today
- route ended meetings from Earlier today into Notes as a manual transcript flow
- create or reuse a manual session for that calendar event
- open the Transcript tab and prompt for Add Transcript only when the session is still empty
- add repository, controller, and grouping coverage for the new path

## Why

Earlier meetings from today could disappear from Coming up once the rest of the day was empty. That made it impossible to use the dashboard as an entry point for adding a transcript to a meeting that had already ended.

## Validation

- `swift test --package-path OpenOats --filter 'NotesControllerTests|SessionRepositoryTests|UpcomingCalendarGroupingTests'`

### DIFF
--- a/OpenOats/Sources/OpenOats/App/AppCoordinator.swift
+++ b/OpenOats/Sources/OpenOats/App/AppCoordinator.swift
@@ -20,6 +20,7 @@ final class AppCoordinator {
         enum Target: Equatable {
             case session(String)
             case meetingHistory(CalendarEvent)
+            case manualTranscript(CalendarEvent)
             case clearSelection
         }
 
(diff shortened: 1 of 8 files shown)


yazins-ai:
Thanks @kkarimi — great feature addition! The earlier-today transcript flow fills a real gap in the dashboard UX. Clean implementation with solid test coverage across the repository, controller, and grouping layers. Merged!

PR #492, opened 2026-04-27
Recover empty cloud live sessions and surface health warnings

### LINKED ISSUE #491: Recover empty cloud live meetings and surface health warnings
## Summary

Cloud live transcription sessions can finish with `0` utterances and leave the user with an empty session stub.

## Root cause

When a cloud live backend fails to produce live text, OpenOats currently depends on the user's recording settings for any recovery path:

- if `Save audio recording` is off
- and `Re-transcribe with higher accuracy after meeting` is off

then no audio is retained and the failed live session is unrecoverable.

The UI also stays too quiet while this is happening, so the user may only discover the failure after the meeting ends.

## Expected

- Cloud live sessions should retain temporary recovery audio automatically.
- If a cloud live session ends with no live utterances, OpenOats should recover it with batch transcription.
- During recording, the UI should surface capture/transcript health warnings early enough for the user to react.

kkarimi:
Fixes #491

## What changed

- keep temporary recovery audio for cloud live transcription sessions even when the user has not enabled audio export or batch retranscription
- auto-run batch recovery when a cloud live session ends with `0` utterances
- surface recording health warnings in the control bar for:
  - no captured audio after startup
  - audio flowing but no live transcript arriving
- mention that cloud recovery batch will run after stop when the stalled live session is using a cloud model

## Why

Cloud live sessions could fail quietly and leave the user with an empty session stub. If the user also had audio export and batch disabled, there was no recovery path and no useful warning during the meeting.

## Validation

- `swift test --filter LiveSessionControllerTests`
- `swift build -c debug`

### DIFF
--- a/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
+++ b/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
@@ -3,6 +3,16 @@ import Observation
 import CoreAudio
 import AppKit
 
+struct RecordingHealthNotice: Equatable {
+    enum Severity: Equatable {
+        case warning
+        case error
+    }
+
+    let severity: Severity
+    let message: String
+}
+
 /// Published state for the live session, projected by ContentView.
 /// Declared as @Observable class so SwiftUI tracks each property individually,
 /// preventing a full view-tree re-render whenever any single field changes.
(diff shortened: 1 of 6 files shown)


yazins-ai:
Thanks @kkarimi — great work on the cloud session recovery and health warnings. The AudioRetentionPlan approach cleanly centralizes the audio-retention logic, and the health notices give users early visibility into capture issues. Merged!

PR #494, opened 2026-04-27
Sort Notes sidebar groups by recency and make them collapsible

### LINKED ISSUE #493: Sort Notes sidebar groups by recency and make them collapsible
## Summary

The Notes sidebar can make recent meetings look missing when sessions are grouped under folders or import-source sections.

## Problem

Two behaviors combine badly:

- section ordering does not reflect recent activity strongly enough
- top-level groups like `My notes` or imported-source sections are not collapsible

That means a newly moved or newly created meeting can end up under the correct folder, but still feel buried because the user is looking at the wrong section and cannot quickly reduce the noise.

## Expected

- sidebar groups should sort by newest session they contain
- `My notes`, folder sections, and source sections should be collapsible
- collapse state should persist locally so the sidebar stays usable across sessions

kkarimi:
Fixes #493

## What changed

- sort Notes sidebar folder sections by the newest session they contain
- let `My notes` participate in that same recency ordering instead of pinning it above more active sections
- sort source sections such as `Granola` by latest activity as well
- make folder and source sections collapsible
- persist the collapse state locally

## Why

Recent meetings could look missing even when they had been filed correctly, because the relevant folder or source section was visually buried and could not be collapsed out of the way. This makes the sidebar reflect current activity instead of static group order.

## Validation

- `swift test --filter NotesControllerTests`
- `swift build -c debug`

### DIFF
--- a/OpenOats/Sources/OpenOats/App/NotesController.swift
+++ b/OpenOats/Sources/OpenOats/App/NotesController.swift
@@ -74,6 +74,7 @@ struct SessionFolderGroup: Identifiable {
     let id: String
     let title: String
     let sessions: [SessionIndex]
+    let isRoot: Bool
 }
 
 struct MeetingFamilySelection: Equatable {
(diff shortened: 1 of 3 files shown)


yazins-ai:
Thanks @kkarimi — nice improvement to the sidebar UX. Recency-based sorting and collapsible sections make the Notes sidebar much more usable when there are many folders and import sources. Merged!

PR #497, opened 2026-04-27
Prompt to move saved meetings from Coming up folder changes

### LINKED ISSUE #496: Prompt to move saved meetings when changing Coming up folder
## Summary

Changing a meeting-family folder from the `Coming up` dashboard only updates the default for future meetings, even when the family already has saved sessions.

## Problem

This makes the `Coming up` folder picker weaker than the equivalent meeting-family folder control in Notes:

- future meetings pick up the new folder
- older saved meetings stay where they were
- the user is not asked whether they want to move those saved meetings too

That is surprising because the control looks like a family-level folder action, not a future-only preference.

## Expected

When a meeting family already has saved meetings, changing its folder from `Coming up` should offer the same choice as the Notes flow:

- `Future meetings only`
- `Move saved meetings too`

kkarimi:
Fixes #496

## What changed

- when changing a meeting-family folder from the `Coming up` dashboard, detect whether that family already has saved meetings
- if it does, show the same choice used in the Notes flow:
  - `Future meetings only`
  - `Move saved meetings too`
- apply the chosen folder to existing saved meetings when requested
- use the same prompt flow for new-folder creation from `Coming up`

## Why

The `Coming up` folder picker looked like a family-level action, but previously only changed the default for future meetings. That left older saved meetings behind with no prompt, which made the behavior feel inconsistent with the Notes meeting-family flow.

## Validation

- `swift test --filter UpcomingCalendarGroupingTests`
- `swift build -c debug`
- `./scripts/run_ui_smoke.sh` was attempted, but the UITest runner timed out while enabling automation mode in the local environment

### DIFF
--- a/OpenOats/Sources/OpenOats/Views/IdleHomeDashboardView.swift
+++ b/OpenOats/Sources/OpenOats/Views/IdleHomeDashboardView.swift
@@ -12,9 +12,17 @@ struct IdleHomeDashboardView: View {
     @State private var showsEarlierToday = false
     @State private var refreshTick = 0
     @State private var creatingFolderEvent: CalendarEvent?
+    @State private var pendingMeetingFamilyFolderChange: PendingMeetingFamilyFolderChange?
     @State private var newFolderPath = ""
     @State private var newFolderColor: NotesFolderColor = .orange
     @FocusState private var newFolderFieldFocused: Bool
+
+    private struct PendingMeetingFamilyFolderChange: Equatable {
+        let historyKey: String
+        let familyTitle: String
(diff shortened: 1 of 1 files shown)


yazins-ai (approved):
Clean implementation. The confirmation dialog correctly checks for existing saved sessions before prompting, handles pluralization, and reuses the existing session repository infrastructure. All three CI checks pass.

yazins-ai:
Merged — thanks @kkarimi for aligning the Coming up folder picker with the Notes flow. The confirmation dialog for existing saved meetings is a nice UX touch.

PR #499, opened 2026-04-27
Add explicit meeting-family move from focused history sessions

### LINKED ISSUE #498: Add explicit meeting-family folder move from focused past sessions
## Summary
When a user focuses a specific past meeting in the meeting history view, the header folder chip currently only exposes single-session moves. That makes it hard to apply the same folder change across the meeting family from the place where the user is already working.

## Desired behavior
- keep the existing per-session move behavior
- add an explicit family-level folder move action from the same focused-session header control
- reuse the existing meeting-family folder/default flow, including prompts about moving saved meetings

## Why
This keeps session moves and family moves both available without silently turning a session-level action into a broader change.

kkarimi:
Fixes #498

## Summary
- keep the focused past-session header folder chip as a session-level move control
- add an explicit `Move meeting family…` submenu from that same control
- reuse the existing meeting-family folder/default flow, including prompts to move saved meetings

## Validation

### DIFF
--- a/OpenOats/Sources/OpenOats/Views/NotesView.swift
+++ b/OpenOats/Sources/OpenOats/Views/NotesView.swift
@@ -1399,66 +1399,13 @@ struct NotesView: View {
         folders: [NotesFolderDefinition]
     ) -> some View {
         Menu {
-            Button {
-                requestMeetingFamilyFolderChange(
-                    controller: controller,
-                    selection: selection,
-                    folderPath: nil,
-                    historyCount: historyCount
-                )
-            } label: {
-                HStack {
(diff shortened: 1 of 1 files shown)


yazins-ai (approved):
Good refactor — extracting the folder menu items into a reusable ViewBuilder and adding the family-level submenu to the session chip is the right approach. Preserves existing session-level behavior while making family moves discoverable from the focused view. All CI checks pass.

yazins-ai:
Merged — thanks @kkarimi for the clean refactor. Extracting the folder menu items into a reusable ViewBuilder and adding the family-level submenu from the session chip was the right call.

PR #509, opened 2026-04-28
Mark empty failed live sessions explicitly

### LINKED ISSUE #508: Mark empty failed live sessions explicitly
## Summary

Some live meetings still end as empty session stubs even though the app looked like it was recording. When that happens, the UI currently falls back to a generic `No transcript` state, which makes real capture/transcription failures look the same as intentionally empty meetings.

## Problems

- empty failed sessions are not classified explicitly
- recovery can succeed later, but the post-session banner can keep showing a stale failure state
- zero-utterance failed sessions can be re-dated incorrectly in history
- Notes/history views do not surface a clear failure reason or the obvious recovery action

## Desired behavior

- classify known empty-session failures explicitly, at least:
  - `No audio captured`
  - `Transcription failed`
- keep intentionally ambiguous empty sessions as plain `No transcript`
- clear the failure marker when a transcript is later recovered
- surface the correct reason and recovery affordance in the main window and Notes/history views

kkarimi:
Fixes #508

## What changed

- persist a `transcriptIssue` marker for live sessions that end empty for a known reason
- distinguish `No audio captured` from `Transcription failed` instead of collapsing both into generic `No transcript`
- show a warning-style post-session banner for failed-empty sessions in the main window
- surface issue-specific copy in Notes/history views and show `Re-transcribe` directly when retained audio exists
- clear the failure marker automatically once a later transcript is saved
- preserve the original meeting start time for zero-utterance sessions
- refresh the last-ended session state when recovery updates the same session after batch completion

## Why

The app could previously finalize an empty live session in a misleading way:
- capture and transcription failures looked the same as intentionally empty meetings
- successful later recovery could leave the main window showing a stale failure outcome
- zero-utterance failures could be re-dated incorrectly in history

This change makes those failed-empty outcomes explicit without auto-deleting ambiguous empty sessions.

## Validation

### DIFF
--- a/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
+++ b/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
@@ -138,6 +138,9 @@ final class LiveSessionController {
                             await notifService.postBatchCompleted(sessionID: sid)
                         }
                         await coordinator.loadHistory()
+                        if coordinator.lastEndedSession?.id == sid {
+                            coordinator.lastEndedSession = await coordinator.sessionRepository.loadSession(id: sid).index
+                        }
 
                         Task { @MainActor in
                             try? await Task.sleep(for: .seconds(3))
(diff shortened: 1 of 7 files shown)


yazins-ai:
Thanks @kkarimi — great work on the failure classification and recovery flow. The distinction between capture-level and transcription-level failures is exactly right, and the automatic clearing on recovery is a nice touch. Merged!

Issue #514, opened 2026-04-28
Improve live recovery UX for failed empty sessions


kkarimi:
## Summary
Improve the user-facing recovery flow after a live session ends empty.

## Scope
- show `Recovery queued` instead of a redundant manual action when automatic recovery is already pending
- add direct `Open Session` / `Re-transcribe` affordances where appropriate
- surface `Recovered after batch` in the home banner, session rows, previous-meeting cards, and transcript detail
- keep ambiguous `No transcript` copy neutral
- clear stale queued-recovery state when the batch engine has moved on to a different session
- clear recovery markers when a transcript is later replaced manually or restored

## Validation
- focused controller, storage, and notes suites
- app bundle build

yazins-ai:
Merged via #515 — thanks @kkarimi for the detailed issue and implementation! The recovery flow now shows "Recovery queued" while batch recovery is pending, surfaces "Recovered after batch" status across the UI, and includes proper "Re-transcribe" affordances and stale-state cleanup.

PR #515, opened 2026-04-28
Improve live recovery UX for empty sessions

kkarimi:
## Summary
- improve the user-facing recovery flow after a live session ends empty
- show `Recovery queued` while automatic recovery is already pending
- surface `Recovered after batch` across the home banner, Notes rows, history cards, and transcript detail

## Details
- add direct `Open Session` and conditional `Re-transcribe` actions where they actually make sense
- keep ambiguous `No transcript` copy neutral instead of implying success or failure
- persist and clear transcript recovery state correctly when transcripts are replaced or restored
- harden queued recovery state so it is cleared when the batch engine has moved on to a different session

## Validation

Related to #514.

### DIFF
--- a/OpenOats/Sources/OpenOats/App/AppCoordinator.swift
+++ b/OpenOats/Sources/OpenOats/App/AppCoordinator.swift
@@ -19,6 +19,7 @@ final class AppCoordinator {
     struct NotesNavigationRequest: Equatable {
         enum Target: Equatable {
             case session(String)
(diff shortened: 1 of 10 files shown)


yazins-ai:
Thanks @kkarimi — clean implementation that follows existing patterns well. The recovery state lifecycle management (tracking active batch session, clearing stale state) and the differentiated UI states (queued vs retranscribable vs nothing) are well done. Merged!

PR #519, opened 2026-04-29
Fix manual transcript reuse across recurring occurrences

kkarimi:
## Summary
- stop manual transcript session reuse from treating a recurring calendar ID as a cross-day exact occurrence match
- create a new manual transcript session when a recurring meeting occurrence is on a different day
- add repository and controller regressions for the `Coming up` -> `Add Transcript` path

## Root Cause
Manual transcript session reuse treated `calendarEvent.id` as an exact occurrence key before checking time proximity. In recurring EventKit data, different occurrences in the same series can share that identifier, so clicking today's meeting from `Coming up` could reopen an older session from the same family.

## Validation

### DIFF
--- a/OpenOats/Sources/OpenOats/Storage/SessionRepository.swift
+++ b/OpenOats/Sources/OpenOats/Storage/SessionRepository.swift
@@ -1161,12 +1161,11 @@ actor SessionRepository {
             }
 
             let gap = abs(metadata.startedAt.timeIntervalSince(referenceDate))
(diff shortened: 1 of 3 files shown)


yazins-ai (approved):
Code review passed. The fix correctly enforces time proximity before calendar event ID matching, preventing recurring meeting occurrences from reusing sessions across different days. Tests cover both the repository and controller layers. CI is green. Merging with a Homebrew cask bump to v1.68.3.

PR #522, opened 2026-04-29
Add cloud live segment diagnostics

kkarimi:
## Summary
- add structured segment-level diagnostics for the cloud live transcription path
- thread the live session ID and active model through the transcription engine into the streaming transcriber
- capture per-segment outcomes (`success`, `empty`, `error`) with normalized cloud/transport error kinds
- add focused tests for structured JSON encoding and cloud error classification

## Why
The recent live cloud failures were no longer pure capture failures. We needed to know whether live segments were succeeding, returning empty text, timing out, hitting transport problems, or being throttled.

## Impact
This does not change core meeting behavior. It improves observability so the live cloud path can be debugged with evidence instead of guesswork.

## Validation
- `swift test --package-path OpenOats --filter TranscriptionEngineTests`

### DIFF
--- a/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
+++ b/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
@@ -523,7 +523,8 @@ final class LiveSessionController {
             await coordinator.transcriptionEngine?.start(
                 locale: settings.locale,
                 inputDeviceID: settings.inputDeviceID,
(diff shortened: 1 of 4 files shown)


yazins-ai:
Merged — thanks @kkarimi for the clean implementation. The structured diagnostics and error classification will help debug cloud transcription issues with evidence instead of guesswork.

PR #526, opened 2026-04-29
Polish prominent button states in settings and setup

kkarimi:
## Summary
- finish moving the remaining settings and setup primary actions onto `OpenOatsProminentButtonStyle`
- refine the shared style so disabled and pressed states have explicit background, foreground, border, and shadow treatment
- keep the existing red and accent variants while making inactive buttons read more deliberately

## Why
Most prominent actions had already moved to the custom style, but a few settings and wizard flows still used `.borderedProminent`. That left inconsistent visuals and weaker disabled-state behavior across the app.

## Validation
- `./scripts/run_ui_smoke.sh` built the macOS UI test host successfully, but execution was blocked by a local environment issue: `Timed out while enabling automation mode`

### DIFF
--- a/OpenOats/Sources/OpenOats/Views/OpenOatsProminentButtonStyle.swift
+++ b/OpenOats/Sources/OpenOats/Views/OpenOatsProminentButtonStyle.swift
@@ -13,20 +13,63 @@ private struct ProminentButtonBody: View {
     let color: Color
 
     @Environment(\.controlSize) private var controlSize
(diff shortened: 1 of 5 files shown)


yazins-ai:
Thanks for the clean work, @kkarimi! The button state refinements look great — explicit disabled/pressed treatments in both color schemes are a solid improvement over the opacity-only approach. Merged via squash.

PR #541, opened 2026-04-30
Preflight broken cloud live starts

### LINKED ISSUE #540: Preflight cloud live starts and surface direct recovery actions
## Summary
Cloud live meetings can still start in obviously broken states, such as a missing API key or stale audio-device selection, and only surface the problem after the user waits through the recording UI.

## Scope
- preflight cloud live starts for missing cloud credentials
- preflight cloud live starts for stale or unavailable microphone/output devices
- surface direct actions from the control bar for deterministic failures
- keep local-model start behavior unchanged

## Why
This is the next reliability step after the live-cloud diagnostics and transcript UX work: do not let a clearly broken cloud meeting start, and give the user an immediate path to fix it.

kkarimi:
Closes #540

## Summary
- preflight cloud live starts for missing API keys and stale microphone/output-device selections
- reuse validated cloud backends on actual start so auth validation is not repeated
- surface direct control-bar actions for deterministic failures, including Settings and microphone privacy shortcuts
- escalate cloud API-key rejections into a blocking error banner instead of only transcript-area copy

## Validation
- `swift test --filter LiveSessionControllerTests`
- `swift test --filter TranscriptionEngineTests`
- `swift build -c debug`

### DIFF
--- a/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
+++ b/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
@@ -114,6 +114,7 @@ final class LiveSessionController {
     private let container: AppContainer
 
     private var downloadTask: Task<Void, Never>?
+    private var startPreflightTask: Task<Void, Never>?
     private var scratchpadSaveTask: Task<Void, Never>?
     private var pendingInitialScratchpad: String?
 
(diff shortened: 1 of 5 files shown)


yazins-ai:
Thanks @kkarimi — solid implementation. The preflight validation with graceful degradation for non-blocking errors and the prepared-backend reuse to avoid double validation are well thought out. Squash-merged.

PR #546, opened 2026-04-30
Add ElevenLabs API key validation

kkarimi:
## What changed

- Added a read-only ElevenLabs API key validator using the same `/v1/voices` endpoint the Scribe backend already uses for preflight validation.
- Added debounced validation state to the Transcription settings key field, showing progress, success, rejected-key, and network-warning states.
- Added focused unit coverage for empty ElevenLabs keys and shared HTTP response mapping.

## Why

Users can now catch broken ElevenLabs credentials when entering the key instead of discovering the problem only when a transcription session starts.

## Validation

- `swift test --package-path OpenOats --filter APIKeyValidatorTests`

### DIFF
--- a/OpenOats/Sources/OpenOats/Views/SettingsView.swift
+++ b/OpenOats/Sources/OpenOats/Views/SettingsView.swift
@@ -338,6 +338,9 @@ private struct TranscriptionSettingsTab: View {
     @Bindable var settings: AppSettings
     @State private var inputDevices: [(id: AudioDeviceID, name: String)] = []
     @State private var outputDevices: [(id: AudioDeviceID, name: String)] = []
+    @State private var isValidatingElevenLabsKey = false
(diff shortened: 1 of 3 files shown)


yazins-ai (approved):
Clean PR — good UX improvement and sensible refactoring.

**What's here:**
- Read-only ElevenLabs key validation via the same `/v1/voices` endpoint already used by Scribe
- Debounced (400ms) validation state in the Transcription settings, with progress/success/error indicators
- Extracted shared `validationResult(for:authFailureMessage:)` to deduplicate OpenRouter/VoyageAI/ElevenLabs HTTP response handling
- Focused unit tests for empty keys and the shared response mapper

**Review notes:**
- No security concerns — key is sent to the expected ElevenLabs endpoint via read-only GET
- Good cancellation handling on view disappear
- Backward-compatible: existing OpenRouter and VoyageAI validation unchanged
- All CI checks pass (validate-swift, package-smoke, ui-smoke)

PR #559, opened 2026-04-30
Improve cloud live responsiveness during chunk processing

### LINKED ISSUE #558: Improve cloud live responsiveness during chunk processing
## Summary
Cloud live transcription currently feels stalled because chunk transcription work blocks the capture/VAD loop.

## This issue covers
- decoupling cloud chunk transcription from the live capture loop
- keeping capture responsive while cloud requests are in flight
- surfacing a clearer in-meeting distinction between waiting for the first chunk and actively processing a chunk

## Realistic progress
This improves responsiveness and status handling. It does **not** fully solve transcript lag under heavy cloud latency. If cloud processing is slower than incoming speech, queued chunks can still accumulate and transcript delay can still grow.

## Follow-up work
- add queue backlog policy so lag does not grow without bound
- evaluate chunk merging or dropping policy when cloud latency is too high
- revisit cloud flush cadence after measuring real latency with the new architecture
- consider backend-specific retry/backoff only after the queue policy is in place

kkarimi:
Closes #558

## Summary
- decouple cloud chunk transcription from the live capture/VAD loop
- keep cloud chunk processing serialized in the background instead of blocking live capture
- show a clearer in-meeting distinction between waiting for the first chunk and actively processing a chunk

## Validation
- swift test --package-path OpenOats --filter StreamingTranscriptionSegmentQueueTests
- swift test --package-path OpenOats --filter TranscriptionEngineTests

## Scope note
This improves responsiveness and status handling. It does **not** fully solve transcript lag when cloud latency is slower than incoming speech. A follow-up should add queue backlog policy and then revisit chunk cadence with real measurements.

### DIFF
--- a/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
+++ b/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
@@ -1065,20 +1065,28 @@ final class LiveSessionController {
 
     static func liveTranscriptNotice(
         for model: TranscriptionModel,
-        issue: CloudTranscriptCopy.Presentation? = nil
+        issue: CloudTranscriptCopy.Presentation? = nil,
+        isProcessing: Bool = false
     ) -> String? {
         if let issue {
             return issue.title
+        }
+        if isProcessing {
+            return CloudTranscriptCopy.processingChunk.title
         }
         return CloudTranscriptCopy.steadyStateNotice(for: model)
     }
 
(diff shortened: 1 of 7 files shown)


yazins-ai:
Merged — thanks @kkarimi for the well-structured contribution! The StreamingTranscriptionSegmentQueue design is clean, and the tests cover the key behaviors well.

PR #561, opened 2026-05-01
fix: prevent ~25 dB voice attenuation on built-in multi-channel mics

Tomdekel:
## Summary

`AudioRecorder.writeMicBuffer` averages all input channels when downmixing to mono. This is correct for stereo music but **wrong for built-in MacBook mic arrays**, which expose 3 beamformed channels: front-facing primary beam (channel 0) plus side and cancellation beams.

Averaging the array causes destructive interference and silently attenuates the recorded voice by ~25 dB. The audio level meter still shows full level (it reads the raw multi-channel buffer pre-downmix), so the bug is invisible at runtime — users only notice when they listen to the saved m4a after a meeting.

## Reproduction

On any modern MacBook (M2/M3 Air, M3 Pro etc., where the built-in mic reports 3 channels):

1. Start a recording with `saveAudioRecording: true`
2. Speak normally for ~30 seconds
3. Open the saved m4a in Quicktime — voice is barely audible (peaks ~-70 dB) despite the level meter showing full activity during recording
4. Compare with `ffmpeg -f avfoundation -i ":0"` directly: same mic, same room, peaks at -40 dB ambient

## Root cause

```swift
// Before — at line 110
let scale = 1.0 / Float(channels)
for ch in 0..<channels { sum += src[ch][i] }
dst[i] = sum * scale
```

For a beamformed mic array where channel 0 is the user's voice and channels 1+ carry directional/cancellation signal that's anti-phase or uncorrelated with channel 0, this averaging:
- Sums anti-phase signals → near-zero
- Divides by N → further attenuation

## Fix

Take channel 0 directly. Channel 0 is the front-facing primary beam, which already carries the user's voice cleanly. For non-interleaved buffers that's a contiguous memcpy; for interleaved buffers we stride by channelCount.

Same logic applied to the int16 and int32 fallback paths.

## Validation

Tested on a MacBook Air (3-channel built-in mic):

| | Before | After |
|---|---|---|
| Peak in mic.caf | -70 dB (~0.0003) | -7 dB (~0.45) |
| Speech recoverable in m4a | barely | clearly |
| Recovery | — | **~63 dB** |

`swift test --filter AudioRecorderTests` passes 8/8 including 3 new regression tests:

- `testInterleavedMultiChannelMicPreservesSignal` — 2-channel anti-phase, interleaved
- `testNonInterleavedMultiChannelMicPreservesSignal` — 2-channel anti-phase, non-interleaved
- `testSingleChannelMicPassThroughUnchanged` — mono path, ensure no regression

The full test suite passes locally except 3 environment-dependent `MeetingDetectorTests` that fail when WhatsApp is running on the test host (unrelated to this change).

## Test plan

- [x] Unit tests added for the regression case (anti-phase channels)
- [x] Unit test added for single-channel pass-through (no regression)
- [x] Manually verified on M-series MacBook Air with WhatsApp self-call: m4a peaks recovered from -70 dB to -7 dB
- [ ] Reviewer to verify on a different mic configuration (Intel Mac, USB mic, AirPods, etc.)

### DIFF
--- a/OpenOats/Sources/OpenOats/Audio/AudioRecorder.swift
+++ b/OpenOats/Sources/OpenOats/Audio/AudioRecorder.swift
@@ -99,65 +99,48 @@ final class AudioRecorder: @unchecked Sendable {
             let dst = monoBuf.floatChannelData?[0] else { return }
             monoBuf.frameLength = buffer.frameLength
 
+            // Multi-channel built-in MacBook mics report 3 channels because
+            // CoreAudio exposes the front-facing primary beam alongside side
+            // and cancellation beams. Channel 0 is the user's voice; channels
+            // 1+ carry directional/cancellation signal that is anti-phase or
+            // uncorrelated with channel 0. Averaging across all channels causes
+            // destructive interference and attenuates the recorded voice by
+            // ~25 dB on a 3-mic array, while the audio level meter (which
+            // reads the raw multi-channel buffer) keeps showing the user's
+            // voice at full level — making the bug invisible at runtime.
+            //
+            // Take channel 0 directly. For non-interleaved buffers that's a
+            // contiguous memcpy of the first channel's plane. For interleaved
+            // buffers we stride by channelCount and grab the first sample of
+            // each frame.
             if let src = buffer.floatChannelData {
-                if channels == 1 {
-                    if buffer.format.isInterleaved {
-                        memcpy(dst, src[0], frames * MemoryLayout<Float>.size)
-                    } else {
-                        memcpy(dst, src[0], frames * MemoryLayout<Float>.size)
+                if buffer.format.isInterleaved {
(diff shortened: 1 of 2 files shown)


yazins-ai (approved):
Thorough and well-tested fix. The root cause analysis is correct — averaging beamformed mic channels causes destructive interference, and channel 0 is the primary beam (user's voice).

Code review notes:
- Float/int16/int32 paths all correctly take ch0 via stride (interleaved) or memcpy/direct-index (non-interleaved)
- Removing the `channels == 1` special case is a nice simplification — ch0-only is a no-op for mono
- Tests cover the anti-phase regression case and single-channel pass-through
- All CI checks green (validate-swift, package-smoke, ui-smoke)

Merging as-is — clean contribution. Thanks @Tomdekel!

PR #572, opened 2026-05-03
Initialize the menu bar controller at app launch

### LINKED ISSUE #571: Initialize the menu bar controller deterministically at app launch
## Problem
The installed app can fail to show its menu bar item even though the same build logic works in the dev app. The current setup ties status-item creation too closely to window appearance and assumes the status-item button is ready during controller init.

## Scope
- initialize the menu bar controller during app launch instead of relying on main-window appearance
- refresh the status item button when the app becomes active instead of assuming it exists only at init time
- add diagnostics breadcrumbs around launch/menu setup so future menu bar regressions are easier to prove

## Non-goals
- do not change the screen-share hiding setting semantics
- do not redesign the app's regular/accessory visibility model in this slice

## Why now
This is blocking confidence in the installed app experience: OpenOats-Dev shows the menu bar item reliably while OpenOats can fail to.

kkarimi:
Closes #571

## Summary
- initialize the menu bar controller during app launch instead of relying on the main window appearance path
- refresh the status item button when the app becomes active instead of assuming it is only available during controller init
- add diagnostics breadcrumbs around menu-bar setup so future launch-path regressions are easier to prove

## Validation

## Notes
- this keeps `hideFromScreenShare` scoped to window sharing only
- this does not redesign the broader regular/accessory activation-policy model in this PR

### DIFF
--- a/OpenOats/Sources/OpenOats/App/MenuBarController.swift
+++ b/OpenOats/Sources/OpenOats/App/MenuBarController.swift
@@ -10,6 +10,7 @@ final class MenuBarController {
     private let settings: AppSettings
     private let onToggleMeeting: () -> Void
     private var iconUpdateTask: Task<Void, Never>?
+    private var hasConfiguredButton = false
 
     var onShowMainWindow: (() -> Void)?
     var onQuitApp: (() -> Void)?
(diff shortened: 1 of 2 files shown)


yazins-ai:
Merged — thanks @kkarimi! The deterministic initialization approach and the `applicationDidBecomeActive` refresh are a solid fix for the installed-vs-dev discrepancy. Diagnostics breadcrumbs will help catch future regressions early.

PR #575, opened 2026-05-03
Add a dedicated Calendar settings tab with filtering

### LINKED ISSUE #574: Add a dedicated Calendar settings tab with visible calendar filtering
## Summary
Move Calendar out of General and make the integration inspectable and filterable.

## Why
- Settings currently say Calendar access is authorized without showing what OpenOats can actually see.
- Bug reports like #570 are harder to diagnose because the user cannot verify which calendars are visible to OpenOats.
- Calendar filtering should be explicit and user-controlled instead of all-or-nothing.

## Scope
- dedicated Calendar settings tab
- visible list of discovered calendars
- include/exclude selection per calendar
- reload control that fully rebuilds CalendarManager / EventKit state
- filtered calendar usage across meeting matching and dashboard previews
- tighter wording and layout so the tab reads as enable -> choose calendars -> optional cloud sharing

## Non-goals
- account/group-level selection UI
- advanced event preview or diagnostics beyond the visible calendar list

kkarimi:
Closes #574

## Summary
- move Calendar out of General into its own Settings tab
- let users include or exclude individual calendars that OpenOats can use
- make reload rebuild CalendarManager / EventKit state so stale visibility bugs are easier to recover from

## UI
- reorganize the tab into a clearer flow: enable Calendar, choose included calendars, optionally share calendar details with cloud notes
- group visible calendars by source/account
- keep reload feedback inside the button instead of adding a separate status row
- tighten wording so local meeting identification and cloud sharing read as different decisions

## Behavior
- persist excluded calendar IDs in settings
- apply the selected calendars to current-event matching, auto-detection, and idle dashboard previews
- prune stale excluded IDs when calendars disappear

## Validation
- `swift test --package-path OpenOats --filter SettingsStoreTests`

### DIFF
--- a/OpenOats/Sources/OpenOats/App/AppContainer.swift
+++ b/OpenOats/Sources/OpenOats/App/AppContainer.swift
@@ -241,7 +241,7 @@ final class AppContainer {
     func updateCalendarIntegration(enabled: Bool) {
         if enabled {
             if calendarManager == nil {
-                calendarManager = CalendarManager()
+                setCalendarManager(CalendarManager())
             } else {
                 // Re-read TCC in case the system state changed since the manager was created.
                 calendarManager?.refreshFromSystem()
@@ -252,8 +252,29 @@ final class AppContainer {
                 }
             }
         } else {
-            calendarManager = nil
+            setCalendarManager(nil)
         }
+    }
+
+    /// Recreates the calendar manager so EventKit state is reloaded from scratch.
+    /// Use this when permission or visible calendars may have changed in System Settings
+    /// and a soft status refresh is not sufficient.
+    func reloadCalendarIntegration() {
+        guard calendarManager != nil else {
+            updateCalendarIntegration(enabled: true)
+            return
+        }
+        setCalendarManager(CalendarManager())
+        if calendarManager?.accessState == .notDetermined {
+            Task {
+                _ = await calendarManager?.requestAccess()
+            }
+        }
+    }
+
+    private func setCalendarManager(_ manager: CalendarManager?) {
+        calendarManager = manager
+        detectionController?.calendarManager = manager
     }
 
     func seedIfNeeded(coordinator: AppCoordinator) async {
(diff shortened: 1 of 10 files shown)


yazins-ai:
Thanks @kkarimi — this is a well-executed implementation of #574. Code review notes below.

**What works well:**
- Clean separation: dedicated `CalendarSettingsView` with card-based layout
- Exclusion-list approach (store what's excluded, include everything by default) is the right default
- `normalizedIdentifierList` in SettingsStore handles deduplication and whitespace properly
- Stale excluded IDs are pruned when calendars disappear
- The `setCalendarManager` helper keeps detection controller in sync
- Tests cover the round-trip and normalization edge cases
- Settings window size bump (500×600 → 640×700) accommodates the new tab

**Observations (non-blocking):**
- The `availableCalendars()` method on CalendarManager returns sorted results which is good for stable UI, but the sort is by source then title — worth confirming this matches what users expect if they have many calendars.
- The 30-second periodic refresh in `.task(id:)` is reasonable for settings visibility but will fire even when the Calendar tab isn't active (as long as the settings window is open). Minor resource consideration.
- `CalendarColorCodec.hexString(from:)` is referenced but not in the diff — assuming this already exists.

**Risk classification:**
- `risk:high` — 10 files across App (4), Meeting (1), Settings (1), Views (3), Tests (1) subsystems. This crosses the 9-file/multi-subsystem threshold per repo policy.
- The risk is procedural rather than substantive — no privacy changes, no new off-device data flow, no consent changes.

Flagging `state:awaiting-human` for maintainer approval on the approach before merge.

yazinsai:
@yazins-ai lgtm; just make sure the `ui-smoke` test passes (might need updating based on this change)

yazins-ai:
Merged — thanks @kkarimi for the clean implementation! The Calendar settings tab with per-calendar filtering is a solid addition. All CI checks (validate-swift, package-smoke, ui-smoke) passed green.

PR #586, opened 2026-05-06
Fix white top and bottom bands in the main window

### LINKED ISSUE #585: Fix white top and bottom bands in the main window
## Summary
When the main window is taller than the material-backed content, OpenOats leaves white bands above and below the main content area.

## Cause
This exposed an older layout bug in `ContentView`: the root content only declared a minimum height, so it did not fill taller windows. The issue became visible after `#578` switched the main window to `.windowResizability(.contentMinSize)`, which stopped collapsing the window back down to the content height.

## Fix
Make the root content fill the available height and stay top-aligned instead of reverting the window-resizability change from `#578`.

## Validation
- `swift build -c debug`

kkarimi:
Closes #585

## Summary
- make the main window content fill the available height instead of only declaring a minimum height
- keep the content top-aligned so taller restored windows do not expose white window background above and below the material content
- fix the latent content-fill bug instead of reverting the `.contentMinSize` window behavior from `#578`

## Why this regressed
The underlying layout bug was already present in `ContentView`, but it became visible after `#578` switched the main window to `.windowResizability(.contentMinSize)`. That change fixed the window collapsing issue, but it also stopped hiding the fact that the content was not filling taller windows.

## Validation
- `swift build -c debug`

### DIFF
--- a/OpenOats/Sources/OpenOats/Views/ContentView.swift
+++ b/OpenOats/Sources/OpenOats/Views/ContentView.swift
@@ -230,6 +230,7 @@ struct ContentView: View {
                 }
             )
         }
+        .frame(maxWidth: .infinity, maxHeight: .infinity, alignment: .top)
     }
 
     private var bodyWithModifiers: some View {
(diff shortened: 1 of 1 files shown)


yazins-ai:
Merged — thanks @kkarimi for the clean fix. Good call on filling the content height rather than reverting the window-resizability change from #578.

PR #590, opened 2026-05-06
Fix system audio capture on USB output devices

andrewphahn:
## Problem

The Core Audio process tap silently produces zero frames whenever the system output is set to a USB audio interface. `AudioHardwareCreateProcessTap`, `AudioHardwareCreateAggregateDevice`, `AudioDeviceCreateIOProcIDWithBlock`, and `AudioDeviceStart` all return `noErr`, but `handleInputData` is never invoked. The same code path works correctly with the MacBook built-in speakers.

Symptoms in the user-facing diagnostics:
- Sessions end with `systemCapturedFrames: false`
- No utterances tagged `[them]` ever appear, only `[you]` from the mic
- "No system audio detected. Check the selected speaker/output device." warning surfaces in the UI
- Switching system output back to MacBook built-in speakers makes capture work again with no other changes

Verified on macOS 26.4.1 against a USB audio interface.

## Root Cause

Two issues compounded:

1. **`tapDescription.stream = 0` was hardcoded.** Many USB audio interfaces expose multiple output streams, or the playback stream is not at index 0 (control/aux channels, multi-stream interfaces). The tap dutifully bound to a stream with no audio activity. Per Apple's docs, `CATapDescription.stream` is `UInt?` and defaults to `nil` = capture from all output streams, which is what we want.

2. **The aggregate device included the tapped USB output device as a real subdevice** (`kAudioAggregateDeviceMainSubDeviceKey` + `kAudioAggregateDeviceSubDeviceListKey`). HAL gets confused trying to drive the aggregate's clock cycle from USB hardware in this topology — the IO proc never fires even though all the setup APIs return `noErr`. Built-in speakers are forgiving of this; USB devices are not. The tap is already pinned to the selected output device via `tapDescription.deviceUID`, so the aggregate's only purpose is to expose the tap as an input source for `AudioDeviceCreateIOProcIDWithBlock` — it doesn't need the device as a subdevice at all.

## Fix

In `OpenOats/Sources/OpenOats/Audio/SystemAudioCapture.swift`:

1. Remove `tapDescription.stream = 0` (let it default to nil = all streams).
2. Switch to a tap-only private aggregate device — drop `kAudioAggregateDeviceMainSubDeviceKey`, `kAudioAggregateDeviceSubDeviceListKey`, and `kAudioAggregateDeviceIsStackedKey`.

The tap remains device-pinned via `tapDescription.deviceUID = outputUID`, so capture is still scoped to the selected output device.

## Verification

- Built locally with `swift build` — clean.
- Tested on macOS 26.4.1 against a USB audio interface:
  - Pre-fix: every test ended with `systemCapturedFrames: false` and zero `[them]` utterances.
  - Post-fix: capture works, IO proc fires, meeting markdown is generated normally with mixed speaker output.
  - macOS re-prompted for "System Audio Recording" permission on the new aggregate topology — expected, since it's a different Core Audio surface.
- MacBook built-in speakers continue to work as before.

## Notes

The "System Audio Recording" TCC prompt may need to be re-granted on first launch with this change because the aggregate device topology is different. Documented behavior — no code change needed.


andrewphahn:
Closing until ready, this s not working yet

andrewphahn:
## Note for local-build testers (TCC gotcha)

When testing this fix from a locally-built debug binary (`swift build`), you need to grant **System Audio Recording** permission to **both**:

1. The OpenOats binary itself, and
2. **The terminal application running the binary** (Terminal.app, iTerm, Warp, etc.)

macOS TCC propagates the System Audio Recording permission from the parent process for unbundled binaries. If the parent terminal doesn't have the permission, macOS silently delivers **zero-filled buffers** to the IO proc rather than failing the Core Audio APIs — every `AudioHardware*` and `AudioDevice*` call still returns `noErr`, frames still arrive on schedule, but every sample is `0.000000`.

This caused a long detour while testing this PR locally — the IO proc would fire on a USB output device but `sysPeak` was always exactly `0.0`, which looked like a deeper Core Audio routing bug. The fix in this PR is correct; granting the terminal the System Audio Recording permission immediately produced non-zero samples and working transcription on the same USB device.

Not a concern for users running the released, signed `.app` — they only need to grant the permission to OpenOats once, as before.

yazins-ai:
Thanks for this, @andrewphahn — the root cause analysis is excellent and the fix is well-targeted.

**Code review:**

The two changes are sound:

1. **Removing `tapDescription.stream = 0`**: Defaulting to `nil` (all output streams) is the correct behavior for diverse hardware. Hardcoding stream index 0 is a silent failure mode on any USB interface where the playback stream isn't at index 0 — exactly the scenario you described.

2. **Tap-only aggregate device**: Dropping the subdevice list (`kAudioAggregateDeviceMainSubDeviceKey`, `kAudioAggregateDeviceSubDeviceListKey`, `kAudioAggregateDeviceIsStackedKey`) is a net simplification. Since the tap is already pinned to the output device via `tapDescription.deviceUID`, the aggregate doesn't need the device as a subdevice — it only needs to expose the tap as an input source. This avoids the HAL clock confusion you identified.

The TCC note in your second comment is very helpful context for local testers — that's a non-obvious macOS behavior.

**One question:** Your earlier comment said "Closing until ready, this is not working yet," but the follow-up TCC explanation suggests the fix is actually correct and the issue was terminal permissions during local testing. Could you confirm whether this PR is ready for review/merge, or if there are still outstanding issues you're working through?

CI checks haven't run on this PR yet — once you confirm readiness we can get those triggered.

andrewphahn:
@yazins-ai PR is ready to merge. It was the TCC explanation. I just needed to grant the terminal system audio recording privileges. Once I found that, I was able to validate that the PR was correct.

yazins-ai:
Thanks for this excellent fix, @andrewphahn! Your root cause analysis was spot on — the hardcoded stream index and the subdevice-based aggregate were both contributing to the silent failure on USB audio interfaces.

I've applied your fix in #591 (reimplemented to run through our CI pipeline since fork PRs can't trigger it directly). Your contribution is credited in the commit message.

The TCC note for local-build testers was also very helpful context. Appreciate you taking the time to document that after the initial debugging detour.

PR #610, opened 2026-05-13
Auto-generate post-meeting notes when configured

### LINKED ISSUE #603: Automatically generate notes (summary) after meeting
It would be great to be able to have an option to automatically generate meeting notes as soon as the meeting ends. We could set a default template or default to the "Generic" one.

My objective is to have the summary + transcript by default in the markdown file right after the meeting ends without having to click.

kkarimi:
Closes #603

## What changed
- trigger automatic note generation after meeting finalization when notes generation is actually configured for the active provider
- wait for batch retranscription when that path is truly active before generating notes
- run background auto-generation through a detached notes engine so it does not clobber a manual generation already running in the detail pane
- refresh the main-window detail pane and the Notes Workspace when background-generated notes land
- add focused provider-gating and lifecycle coverage, plus harden the two home-timeline smoke tests around restored secondary windows

## Why
The current flow requires a manual `Generate Notes` action even when the user has already configured notes generation. That is the behavior gap in #603.

The implementation uses a stricter gate than "a model string exists" because `OpenRouter` ships with a default model name even when no API key is present. Treating that as configured would create post-meeting failures after every meeting.

## User impact
- stopping a meeting can now produce notes automatically without an extra click
- batch-improved transcripts still win when batch retranscription is enabled and available
- existing manual note generation remains intact for users who want to regenerate or edit from the detail pane

## Validation
- `swift test --filter "LiveSessionControllerTests|SettingsStoreTests|NotesControllerTests|NotesEngineTests"`
- `scripts/run_ui_smoke.sh`
- `SKIP_SIGN=1 SKIP_INSTALL=1 scripts/build_swift_app.sh`

### DIFF
--- a/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
+++ b/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
@@ -146,6 +146,8 @@ final class LiveSessionController {
     private var observedSystemHasEverCapturedFrames = false
     private var observedMicHasEverCapturedFrames = false
     private var pendingRecoveryDiagnostics: PendingRecoveryDiagnostics?
+    private var pendingAutoNotesSessionID: String?
+    private var autoGeneratingNotesSessionID: String?
 
     init(coordinator: AppCoordinator, container: AppContainer) {
         self.coordinator = coordinator
@@ -261,6 +263,8 @@ final class LiveSessionController {
                             }
                         }
                     }
+
+                    resumePendingAutoNotesIfNeeded(for: status, settings: settings)
                 }
             }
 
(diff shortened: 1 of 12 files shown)


allaire:
@kkarimi Thanks for this, it works well, but curious to know how to select a default custom template?

kkarimi:
@allaire its meant to be at the meeting "family" level but I think its not working right now? will try to test it out/get a fix in soon

allaire:
@kkarimi how is the family determined? It fallback on Generic so far for me, but perhaps I<m not understand this correctly.

Also unclear why there is a lock and reset button in "Settings > Templates" that seems to do nothing

kkarimi:
@allaire meeting family is meant to be reoccurring calendar events that have the same prefixed meeting id, with fallback of meetings that have same title. 
please do keep posting about any issues you find, I’ll make sure to address all of them, thanks

allaire:
@kkarimi ah, got it. perhaps being able to set a "default" template would be a better UX.

Also noticed that when you generate notes, it re-writes the meeting filename as well, but I think keeping the calendar event name is always better when a recording is linked to a meeting.

PR #635, opened 2026-05-22
fix: Expand Advanced Detection Settings tap target to full row

jasnoorgill:
## Summary

The **Advanced Detection Settings** row in Settings → General had a tap target limited to just the chevron (>). This PR makes the entire light-colored row rectangle tappable.

**Changes:**
- Bound `DisclosureGroup` expansion to a `@State` variable (`showAdvancedDetection`) for manual control
- Replaced the default string label with a custom label using a full-width `HStack + Spacer`
- Used `.padding(.vertical, 10)` → `.contentShape(Rectangle())` → `.padding(.vertical, -10)` to extend the hit area into the Form row's vertical insets, covering the full background rectangle without affecting row height

This is the same pattern used in #611 (intent card tap target fix).

## How to test

1. Open Settings → General
2. Enable **Auto-detect meetings** to reveal the Advanced Detection Settings row
3. Tap anywhere on the light-colored row (top edge, bottom edge, empty space) — it should expand/collapse

### DIFF
--- a/Casks/openoats.rb
+++ b/Casks/openoats.rb
@@ -1,6 +1,6 @@
 cask "openoats" do
-  version "1.79.1"
-  sha256 "d64cb2243fe14a9501859df3c69965b1dede8c6d9d4375fdf70a76283fa6c90e"
+  version "1.80.0"
+  sha256 "0db36f361a9061b3e4b8e23bc33693e667ee9e7043e250103873f7c9deab505a"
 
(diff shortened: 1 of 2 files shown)


yazins-ai:
Thanks @jasnoorgill. The full-row tap target is a nice polish fix for Settings, and I folded in the pending cask bump before merging.

PR #636, opened 2026-05-23
fix: Remove duplicate chevron in OllamaModelField menu

jasnoorgill:
## Summary

The Ollama model selector dropdown was showing two chevrons side by side. The `borderlessButton` menu style automatically renders its own indicator arrow, but the label already contained an explicit `chevron.down` icon — causing the duplication.

**Fix:** Added `.menuIndicator(.hidden)` to suppress the automatic indicator, leaving only the intentional custom icon.

Since `OllamaModelField` is a shared component, this fixes the double chevron in all three places it appears:
- Settings → Intelligence → Notes Generation → Model
- Settings → Intelligence → Knowledge Base Retrieval → Model
- Settings → Intelligence → Classic Suggestions → Model

One-line change in `OllamaModelField.swift`.

### DIFF
--- a/OpenOats/Sources/OpenOats/Views/OllamaModelField.swift
+++ b/OpenOats/Sources/OpenOats/Views/OllamaModelField.swift
@@ -40,6 +40,7 @@ struct OllamaModelField: View {
                         .frame(width: 16, height: 16)
                 }
                 .menuStyle(.borderlessButton)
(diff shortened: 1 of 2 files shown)


yazins-ai:
Thanks @jasnoorgill. The shared Ollama model picker fix is a clean one-liner and removes the double-chevron glitch everywhere that component is used.

PR #637, opened 2026-05-23
fix: Fix template creation form alignment and growth

jasnoorgill:
## Summary

The new template creation form (Settings → Templates → New Template) had two layout bugs caused by being nested inside a `Form { .formStyle(.grouped) }`:

- **Text field right-aligned**: macOS grouped Form applies trailing column alignment to all content, and SwiftUI modifiers like `.multilineTextAlignment(.leading)` and `.environment(\.multilineTextAlignment, .leading)` cannot override it from within
- **Notes Prompt box grew with content**: Form row height management ignored `.frame(height:)` and `.clipped()` constraints

**Fix**: Moved the creation `VStack` out of the `Form` entirely, rendering it directly in the `ScrollView`. This removes Form's layout engine from the picture entirely.

**Additional improvements while here:**
- Upgraded Name / Icon / Notes Prompt labels to `size 13, semibold` for clearer visual hierarchy
- Moved `"e.g. Sprint Planning"` hint inline with the Name label instead of as a separate row
- Reduced text field width slightly with `Spacer(minLength: 20)`
- Replaced `TextEditor` with a custom `FixedLeftTextEditor` (`NSViewRepresentable`) that sets `NSTextView.alignment = .left` directly and implements `sizeThatFits` to pin height at 100pt regardless of content

## How to test

1. Open Settings → Templates
2. Click **New Template**
3. Type in the Name field — text should be left-aligned and the row should stay single-line height
4. Type a long prompt in Notes Prompt — the box should stay fixed at 100pt and scroll internally

### DIFF
--- a/OpenOats/Sources/OpenOats/Views/OllamaModelField.swift
+++ b/OpenOats/Sources/OpenOats/Views/OllamaModelField.swift
@@ -40,6 +40,7 @@ struct OllamaModelField: View {
                         .frame(width: 16, height: 16)
                 }
                 .menuStyle(.borderlessButton)
+                .menuIndicator(.hidden)
                 .fixedSize()
             }
         }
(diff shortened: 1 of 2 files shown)


yazins-ai:
Thanks @jasnoorgill. Moving the template editor out of the grouped Form fixes the alignment and prompt growth issues cleanly, and the fixed-height text editor keeps the form stable.

PR #639, opened 2026-05-23
fix: Match Clean Up button height to Manage and copy buttons

jasnoorgill:
## Summary

The **Clean Up** button was taller than the adjacent **Manage** and copy icon buttons in the transcript toolbar, making the row visually inconsistent.

**Root cause**: `OpenOatsProminentButtonStyle` defaults to `.regular` `controlSize`, which applies 7pt vertical padding on each side. The Manage and copy buttons use the native `.bordered` style which renders at standard macOS control height — shorter than the custom style at `.regular`.

**Fix**: Added `.controlSize(.mini)` to both Clean Up button instances (`.notCleaned` and `.partiallyCleaned` states), reducing vertical padding to 4pt and aligning the button height with the rest of the toolbar.

The blue accent color already gives Clean Up sufficient visual prominence — the extra height was unnecessary.


yazins-ai:
Thanks @jasnoorgill. This cleans up the transcript toolbar nicely by making the Clean Up action align with the Manage and copy controls.

PR #675, opened 2026-06-29
Fix meeting auto-stop never triggering on a live mic (adaptive silence floor)

bigfish24:
## Problem

The feature that auto-stops a recording after a meeting ends never fired for manual (press-Start) sessions — recordings would run for hours or days after the meeting was over.

**Root cause:** the last-ditch silence auto-stop in `LiveSessionController` compared the combined audio level against a **fixed `0.01` threshold**. Because mic/system levels are computed as `min(rms * 25, 1.0)`, that threshold corresponds to roughly **−68 dBFS** — below the noise floor of essentially every microphone. A live mic's ambient/electrical noise (measured **~0.08–0.10 / ≈−50 dBFS** on the reporter's hardware) always exceeded the bar, so every poll reset the silence timer and the timeout never elapsed.

Separately, the user-facing **"Silence timeout"** setting applied only to *auto-detected* sessions and was minutes-only, so it didn't cover manual recordings.

## Solution

**Adaptive, per-mic noise floor.** Replace the fixed threshold with a learned ambient floor:
- An asymmetric EMA tracks the noise floor — falls fast toward quiet (τ≈5s), rises slowly (τ≈60s) so speech bursts don't inflate it.
- Audio counts as *activity* only when it exceeds the floor by **2.5×** (~8 dB headroom). Speech sits 15–40 dB above the floor, so it clears easily; steady mic/room/electrical noise does not.
- Calibrates to any mic automatically — no magic absolute number.

**Unified, configurable timeout.**
- The adaptive stop already runs for *every* session in the polling loop, so wiring its timeout to the setting makes the setting apply to **manual and auto-detected** sessions alike. Resolution order: `OPENOATS_SILENCE_TIMEOUT_SECONDS` env (testing) → user setting → default. `0` disables.
- The setting is now stored in **seconds** with a **sec/min** unit picker in a new always-visible **"Auto-Stop on Silence"** section. The legacy `silenceTimeoutMinutes` value is migrated automatically.
- Removed the redundant, coarse (60s-poll) utterance-based manual monitor; the adaptive audio-level stop replaces it.

**Setting takes effect live (no restart).** The timeout is resolved from the *persisted* value (`SettingsStore.persistedSilenceTimeoutSeconds`) rather than the in-memory cache, so a change made in the Settings window is honored by an in-progress recording on the next poll.

> Note / follow-up: the reason the in-memory cache could go stale is that the live session and the Settings window can hold **different `SettingsStore` instances** (app bootstrap can mint more than one). Reading the persisted value sidesteps that here. A separate change to share a **single** `SettingsStore` instance would make *all* settings update live via `@Observable` and let this shim be removed — intentionally left out of this PR to keep the app-wide bootstrap change reviewable on its own.

### Files
- `LiveSessionController.swift` — adaptive floor (`SilenceTracking`, `updatedNoiseFloor`, rewritten `automaticSilenceTimeoutEvaluation`); setting-driven timeout resolved from the persisted value; removed manual utterance monitor; os_log at auto-stop.
- `SettingsStore.swift` — `silenceTimeoutSeconds` + display-unit preference + migration + `persistedSilenceTimeoutSeconds`.
- `SettingsView.swift` — "Auto-Stop on Silence" section with sec/min picker.
- `MeetingDetectionController.swift` — auto-detected monitor ported to seconds with finer poll cadence.

## Validation

**Unit tests** (`swift test` — all passing):
- `testSteadyMicNoiseFloorStillAutoStops` — constant `0.06` ambient (6× the old threshold) now auto-stops around the timeout instead of never. Direct regression test for the bug.
- `testLoudConstantNoiseFloorAlsoAutoStops` — a noisy `0.30` mic also stops (proves it's relative, not absolute).
- `testPeriodicSpeechAboveFloorPreventsAutoStop` — periodic speech over 15 min never false-stops an active meeting.
- `testNoiseFloorFallsFasterThanItRises` — EMA asymmetry.
- `testPersistedSilenceTimeoutReflectsExternalWrite` — persisted read reflects a live change while the in-memory cache stays stale (covers the no-restart fix).
- Settings: seconds default, roundtrip, negative clamp, **legacy-minutes migration**, unit roundtrip.

**Live run:** built and signed the app and exercised it end-to-end:
- With a short timeout, a manual recording auto-stopped after going silent — confirmed via the log:

  ```
  [com.openoats.app:MeetingDetection] Auto-stopping after Ns of silence (noise floor 0.082105)
  ```

  The logged noise floor (`0.082` ≈ −50 dBFS) is the smoking gun: ~8× the old `0.01` threshold, which is exactly why the old logic never stopped.
- Reproduced and verified the no-restart fix: changing the timeout in Settings is honored by the running session without relaunching.

### DIFF
--- a/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
+++ b/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
@@ -144,7 +144,9 @@ final class LiveSessionController {
            let seconds = TimeInterval(raw), seconds > 0 {
             return seconds
         }
-        if let configured = settings?.silenceTimeoutSeconds {
+        // Read the persisted value so a live recording honors a Settings change
+        // immediately (no app restart), falling back to the in-memory value.
+        if let configured = settings?.persistedSilenceTimeoutSeconds ?? settings?.silenceTimeoutSeconds {
             return TimeInterval(configured)   // 0 == disabled
         }
         return automaticSilenceTimeoutInterval
--- a/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
+++ b/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
@@ -115,13 +115,63 @@ final class LiveSessionController {
         let hasBlockingError: Bool
     }
 
+    /// Persisted state for adaptive silence detection, threaded between
+    /// successive evaluations of the polling loop.
+    struct SilenceTracking: Equatable {
+        var lastAudibleActivityAt: Date?
+        /// Running estimate of the ambient (combined mic + system) audio level.
+        var noiseFloor: Float
+        /// Timestamp of the last sample folded into the floor estimate.
+        var lastSampleAt: Date?
+
+        static let initial = SilenceTracking(lastAudibleActivityAt: nil, noiseFloor: 0, lastSampleAt: nil)
+    }
+
     struct AutomaticSilenceTimeoutEvaluation: Equatable {
-        let lastAudibleActivityAt: Date?
+        let tracking: SilenceTracking
         let shouldStop: Bool
     }
 
     static let automaticSilenceTimeoutInterval: TimeInterval = 300
+
+    /// Effective silence timeout in seconds, applied to every session (manual
+    /// and auto-detected). Resolution order:
+    /// 1. `OPENOATS_SILENCE_TIMEOUT_SECONDS` env override (for testing),
+    /// 2. the user's `silenceTimeoutSeconds` setting (0 disables auto-stop),
+    /// 3. the built-in default.
+    static func effectiveAutomaticSilenceTimeoutInterval(settings: AppSettings?) -> TimeInterval {
+        if let raw = ProcessInfo.processInfo.environment["OPENOATS_SILENCE_TIMEOUT_SECONDS"],
+           let seconds = TimeInterval(raw), seconds > 0 {
+            return seconds
+        }
+        if let configured = settings?.silenceTimeoutSeconds {
+            return TimeInterval(configured)   // 0 == disabled
+        }
+        return automaticSilenceTimeoutInterval
+    }
+
+    /// Absolute lower bound on what counts as "audible". The adaptive floor
+    /// does the real work; this only guards against the dynamic threshold
+    /// collapsing toward zero on a near-silent input. ~ -68 dBFS combined.
     static let audibleActivityLevelThreshold: Float = 0.01
+
+    /// A sample counts as audible activity when it exceeds the estimated
+    /// noise floor by this factor (≈ 8 dB of headroom above ambient). Speech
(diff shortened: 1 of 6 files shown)


bigfish24:
### Note for maintainers: the live-update fix is a targeted workaround — a single shared `SettingsStore` would be the cleaner fix

One commit here (`Apply silence timeout live, without an app restart`) deserves a closer look, because it papers over what looks like a broader architectural issue.

**What this PR does:** resolves the silence timeout from the *persisted* value (`SettingsStore.persistedSilenceTimeoutSeconds`, which re-reads `UserDefaults`) instead of the in-memory cached property, so an in-progress recording honors a Settings change on the next poll.

**Why it was needed:** before this, changing the timeout only took effect after relaunching the app. A running recording kept using the value loaded at launch — across multiple start/stop cycles in the same process — and only a restart picked up the new value. That pattern means the recording session and the Settings window are reading **different `SettingsStore` instances**: the UI's setter updates instance A (and persists), while the session reads instance B, whose in-memory cache was populated at launch and never saw the change. Reading the persisted value (the source of truth, written on every change, on the shared `UserDefaults`) sidesteps the stale cache.

**Why it's a smell:**
- It bypasses the `@Observable` `SettingsStore` abstraction and reads raw `UserDefaults` for one setting, duplicating the key string.
- It only fixes *this* setting. Any other setting read continuously mid-session would have the same staleness; most settings happen to be read once at session start, which is why this one (read every poll) surfaced it.
- It treats the symptom, not the cause.

**The better, broader fix — one shared `SettingsStore`:** if every consumer shares a single instance, `@Observable` + reference semantics make *all* settings update live everywhere for free, and `persistedSilenceTimeoutSeconds` can be deleted.

The duplication appears to originate in app bootstrap:
- `OpenOatsApp.init()` calls `AppContainer.bootstrap()`, which mints `AppSettings()` (`AppContainer.swift`).
- The result is stored both in SwiftUI `@State` (used by `ContentView`/`SettingsView`) **and** in the global `AppLaunchBootstrap.context` (`OpenOatsApp.swift`), which the `AppDelegate` later consumes to configure the menu bar / coordinator.
- If `App.init()` runs more than once (SwiftUI can re-init the `App` value), `bootstrap()` mints a second `AppSettings()`, and the `@State` side and the `AppDelegate`/coordinator side can end up on different instances.

I inferred this from behavior (stale-until-restart across multiple recordings in one process), not from a definitive trace, so the exact path is worth confirming. A likely clean fix is to **memoize the bootstrap context** (or otherwise guarantee a single `AppSettings` per process) so all consumers share it — then revert the persisted-read shim.

I deliberately kept that out of this PR: it touches app-wide bootstrap (onboarding, menu bar, coordinator wiring) and deserves its own review rather than riding along with a meeting-auto-stop fix. Happy to open a follow-up PR for it if that's the preferred direction.

PR #676, opened 2026-07-02
Fix audio-pipeline crashes and data loss on device switches

changbenjamin:
## What changed

Fixes 19 verified bugs in the audio capture/save pipeline, centered on the two user-visible failures: **crashes when switching audio devices** (speakers ↔ AirPods ↔ Bluetooth) and **sessions losing their audio**.

### Crash fixes
- **MicCapture**: the tap was installed at the *hardware* nominal sample rate even when it differed from the engine's input-node format — AVFoundation raises an uncatchable NSException for that mismatch, which is exactly what happens on AirPods/Bluetooth switches. The tap now always uses the node's own rate; mistagged rates are corrected downstream (transcriber cadence detection + stem rebuilder). This bug is inherited from upstream OpenOats.
- **TimingAwareStemRebuilder**: conversion failures used to fall back to writing a format-mismatched buffer into the output `AVAudioFile` (second NSException path, hit during finalization of device-switch sessions). Unconvertible chunks are now skipped, preserving the timeline.
- **diagLog / SessionStore**: legacy `FileHandle.seekToEndOfFile()/write(_:)` raise NSExceptions on write failure (disk full) and were called from audio render callbacks. Rewritten with throwing APIs on a serial queue.
- **SystemAudioCapture**: start/stop are serialized under a control lock with take-and-clear state (concurrent stop + device-switch restart could double-destroy CoreAudio objects); the sample-rate resolver is now confined to the delivery queue (torn-pointer risk).

### Audio-loss fixes
- **Stem truncation (critical)**: a buffer arriving just after `finish()` lazily re-created the `.caf` via `AVAudioFile(forWriting:)`, which truncates — replacing the whole recording with one ~85 ms buffer. `PCMFileWriter` now carries an `isFinished` guard; on drain timeout the writers are force-finished so stragglers can't touch the file.
- **Stop/start re-entrancy (critical)**: starting a new session while the previous one finalized let the suspended `stop()` tear down the *new* session's capture and recorder (shared `SessionAudioRecorder` instance). Engine stops are serialized via `stopTask`, `startSession` awaits pending teardown, and both start/stop are gated in the UI.
- **AirPods profile flips (A2DP↔HFP)**: the rate resolver now keeps monitoring delivery cadence after locking its format and re-resolves on sustained divergence, so the remainder of the "Them" stem is no longer written at the wrong speed. The mixer also refuses to apply a session-average resample to a mixed-rate stem (which corrupted the correctly-rated half too) and defers to the per-chunk rebuilder.
- **Tail loss**: system capture flushed its buffered tail *after* closing the stream, dropping ~0.5 s at every stop and device switch. Flush now happens first.
- **Timing/stem desync**: timing chunks are recorded only after the corresponding write succeeds, and the mixer clamps chunk ranges to the actual file length + skips per-chunk insert failures — a single failed write used to abort the entire combined-audio export.
- **Crash recovery**: the timing JSON is checkpointed periodically during recording (interval grows with session length to bound cost), and on launch the app rebuilds combined audio for any session with stems but no mix (once per process).
- **Long sessions**: the resample path streamed in 64K-frame blocks instead of loading the entire stem into memory ~3× (multi-GB allocations during finalization of 2h+ sessions).
- **Mic failure fallback**: if the selected input device fails (unplugged, dead battery), the engine surfaces an error and falls back to the automatic device instead of silently recording no mic audio.
- **Converter quality**: `PCMFileWriter` keeps a streaming `AVAudioConverter` across buffers (`.noDataNow`) instead of resetting per buffer (audible click ~every 85 ms in rate-converted stems), with tails drained at finish; the rebuild threshold drops to 8% so 44.1/48 kHz mistags are corrected at mix time.

## Why

Users hit intermittent crashes mid-recording and sessions whose audio was missing or wrong-speed, specifically when switching between speakers, AirPods, and Bluetooth headphones. Every fix above corresponds to a bug confirmed by adversarial multi-agent review of the pipeline (19 confirmed findings across capture, storage, and mixing; 5 candidate findings were refuted and not acted on).

## Reviewer notes

- `swift build` (debug + release) passes; the release app was built and smoke-launched via `scripts/build_swift_app.sh`.
- The changes were themselves re-reviewed adversarially; the 6 issues that review surfaced (converter tail drains, drain-timeout stragglers, checkpoint O(n²) cost, launch-recovery double-run, 44.1/48 mistag gap) are fixed in this diff.
- Known residual limitation: a mic mistag within ~8% of the true rate with no device-change notification still relies on cadence inference at mix time; raw stems remain intact and repairable in all cases.
- The `restartMic` / `performSystemCaptureRestart` flow is the most concurrency-sensitive area — worth a close read of the new `isRunning`/cancellation re-checks after each `await`.


changbenjamin:
Opened against the wrong repo (intended for my fork) — apologies for the noise.

PR #688, opened 2026-07-20
Fix cloud credentials for audio imports

gthay:
Fix for Issue: #689

## Summary

- pass the configured cloud ASR API key into imported-recording transcription
- preserve custom vocabulary and filler-word settings for imported recordings

## Root cause

`BatchAudioTranscriber.runImport` created the selected backend with `model.makeBackend()` and therefore used the default empty API key. Cloud backends failed during preparation even when the same saved credentials worked for live meetings.

The import entry point now snapshots the active transcription settings and passes them through to backend construction.

## User impact

Users can import meeting recordings with AssemblyAI and other configured cloud transcription backends without receiving a misleading invalid API key error.

## Validation

- `swift build`
- `git diff --check`
- `swift test` compiled the application and changed sources, but the test target could not run in this environment because the active Command Line Tools installation does not provide the `XCTest` module

### DIFF
--- a/OpenOats/Sources/OpenOats/App/OpenOatsApp.swift
+++ b/OpenOats/Sources/OpenOats/App/OpenOatsApp.swift
@@ -215,6 +215,9 @@ extension OpenOatsRootApp {
 
         let model = settings.transcriptionModel
         let locale = settings.locale
+        let customVocabulary = settings.transcriptionCustomVocabulary
+        let apiKey = settings.cloudASRApiKey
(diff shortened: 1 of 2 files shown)


gthay:
Hi, @yazinsai 
Did you have a chance to look at this PR yet?

yazins-ai:
Thanks @gthay for fixing the cloud credentials handling for audio imports — merged. 🎉

PR #692, opened 2026-08-01
Use AssemblyAI's latest default speech model

gthay:
## Summary

- remove the explicit `universal-3-pro` / `universal-2` model list from AssemblyAI transcript requests
- let AssemblyAI automatically select its latest default speech recognition model
- add a regression test that verifies transcript payloads omit `speech_models`

## Why

AssemblyAI will deprecate `universal-3-pro` on September 2, 2026, after which requests that explicitly name it will fail. The backend currently pins that model in every transcript request.

Because `speech_models` is optional, omitting it lets AssemblyAI route requests to its current Universal default. This avoids the immediate deprecation failure and future app changes solely for renamed default models. See AssemblyAI's [model selection documentation](https://www.assemblyai.com/docs/pre-recorded-audio/select-the-speech-model).

## Impact

AssemblyAI cloud transcription continues working after the deprecation and automatically follows AssemblyAI's latest default model. Existing API keys, language hints, and custom spelling behavior are unchanged.

## Validation

- `git diff --check`
- verified production sources contain neither the deprecated model ID nor an emitted `"speech_models"` payload key
- added `testAssemblyAITranscriptRequestUsesLatestDefaultModel`
- `swift test` could not run in the local Linux workspace because the Swift toolchain is unavailable; the repository's macOS CI will run the full builds and test suite

### DIFF
--- a/OpenOats/Sources/OpenOats/Transcription/AssemblyAIBackend.swift
+++ b/OpenOats/Sources/OpenOats/Transcription/AssemblyAIBackend.swift
@@ -178,23 +178,10 @@ final class AssemblyAIBackend: TranscriptionBackend, @unchecked Sendable {
         }
     }
 
-    // MARK: - Private: Create Transcript
+    // MARK: - Create Transcript
 
     private func createTranscript(audioURL: URL, locale: Locale) async throws -> String {
-        var body: [String: Any] = [
-            "audio_url": audioURL.absoluteString,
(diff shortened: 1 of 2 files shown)


gthay:
Hi @yazinsai,
please review this PR carefully, it's generated with Sol 5.6 and merge.

yazins-ai:
Thanks @gthay for updating the AssemblyAI integration to use their latest default speech model — merged. 🎉

PR #696, opened 2026-08-26
Fix text selection: allow selecting across transcript lines and notes paragraphs (drag + Cmd+A)

Amir-Abushanab:
## Bug

Text selection in the transcript and notes views only works one line/paragraph at a time:

- Dragging the cursor across multiple transcript lines doesn't extend the selection
- Cmd+A doesn't select anything
- Same for paragraphs in the generated notes

**Cause:** each transcript row and notes block is its own SwiftUI `Text` with `.textSelection(.enabled)`, and SwiftUI on macOS can't select across separate `Text` views.

## Fix

New `SelectableTextFlow` view (`Views/SelectableTextFlow.swift`): a non-editable, selectable `NSTextView` that renders the whole content as one text flow, so selection behaves like native macOS text — drag across everything, Cmd+A, Cmd+C, right-click → Copy.

Wired into the three affected surfaces:

- **Live transcript** (`TranscriptView`): the utterance list renders as one flow. The timestamp / speaker / text column layout is reproduced with tab stops (right-aligned timestamp at 34pt, right-aligned speaker column sized to the widest label, hanging indent for wrapped lines). Auto-scroll now targets a bottom anchor instead of per-row IDs; search filtering and the volatile (in-progress) indicators are unchanged, with the volatile rows aligned to the same speaker column width.
- **Session transcript** (`MeetingDetailPane`): same flow rendering. Speaker renaming is preserved — renameable speaker labels are links (`openoats-rename://` scheme, styled identically, hand cursor) that open the existing `SpeakerRenamePopover`, anchored at the clicked label via the link run's bounding rect. Cleanup styling (secondary color while cleaning) and the original/cleaned toggle behave as before.
- **Notes** (`MeetingDetailPane`): headings and markdown text blocks merge into continuous selectable flows (inline bold/italic/code/links mapped to `NSAttributedString`); image and attachment previews stay as interactive cards between flows. Relative asset links keep resolving against the session directory. Copied text preserves blank lines between blocks.

Selection is preserved across live-transcript updates (ranges are restored after the text storage is replaced; new text appends at the end so offsets stay valid).

## Testing

- `swift build -c debug` clean (no new warnings)
- Manually verified in the app: drag-selection across many transcript lines, Cmd+A / Cmd+C in live transcript, session transcript, and notes; speaker rename popover still opens from the label; auto-scroll and search still work

### DIFF
--- a/OpenOats/Sources/OpenOats/Views/MeetingDetailPane.swift
+++ b/OpenOats/Sources/OpenOats/Views/MeetingDetailPane.swift
@@ -60,6 +60,7 @@ struct MeetingDetailPane<SessionFolderMenuItems: View>: View {
     @FocusState private var renameFieldFocused: Bool
     @State private var renamingSpeakerKey: String? = nil
     @State private var speakerRenameText: String = ""
+    @State private var speakerRenameAnchorRect: CGRect = .zero
     @State private var editingTagsSessionID: String?
     @State private var editingTags: [String] = []
     @State private var newTagText: String = ""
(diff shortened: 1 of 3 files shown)


yazins-ai:
Review (autonomous maintainer):

**Deception check:** clean.

**Assessment:** Solid, well-scoped fix. The root cause (per-`Text` `.textSelection(.enabled)` can't select across separate SwiftUI views on macOS) is correct, and routing transcript + notes through a single non-editable `NSTextView` (`SelectableTextFlow`) is the right approach. Changes are confined to the view layer — no data-model, persistence, network, or security surface. Column layout is reproduced via tab stops, speaker rename is preserved as `openoats-rename://` link runs with rect-anchored popover, and note image/attachment cards stay interactive between flows.

Notes (non-blocking):
- `TranscriptView` and the session transcript switch from `LazyVStack` to a single `VStack` + one `NSTextView`. TextKit handles large text well, so this is fine, but very long session transcripts now build one big `NSAttributedString` on each change — acceptable given `updateNSView` short-circuits on `isEqual`.
- Auto-scroll now targets a single `transcript-bottom` anchor instead of per-row IDs — simpler and equivalent for the append-only case.

**Labels:** kind:bug, risk:medium, release:patch.

**state:ready-to-merge** — pending the required checks (`validate-swift`, `package-smoke`, `ui-smoke`), which I've just approved to run. Will squash-merge once they're green.

yazins-ai:
Merged in 31eeca8. Thanks @Amir-Abushanab — this is a clean fix for a long-standing annoyance: replacing the per-row SwiftUI `Text` + `.textSelection` with a single `SelectableTextFlow` NSTextView so selection finally spans the whole transcript and notes (drag, Cmd+A, Cmd+C), while keeping speaker-rename links and inline note attachments working. All three checks (validate-swift, package-smoke, ui-smoke) green. Will go out in the next release.

PR #699, opened 2026-08-27
Add Export to .txt/.pdf/.docx in the session detail toolbar

Amir-Abushanab:
Follow-up to #696.

## What

An **Export** menu (share icon, next to the Copy button) in the session detail toolbar, offering **Plain Text (.txt)**, **PDF (.pdf)**, and **Word (.docx)**. It exports whatever the current tab shows:

- **Transcript** — `[HH:mm:ss] Speaker: text` in plain text; the PDF/Word version keeps the timestamp column, colored speaker labels, and hanging indents (custom speaker names respected)
- **Notes** — raw markdown in plain text; headings and inline styling in PDF/Word (image/attachment blocks become `[Image: …]` / `[Attachment: …]` placeholder lines)
- **Ask** — the conversation with You/OpenOats headings

A save panel proposes a name like `Meeting 2026-08-27 - Transcript.pdf` from the session title and date; the button is disabled when the tab is empty.

## How

- `ContentExporter` (new): save panel + three writers. Plain text shares the exact string the Copy button produces (extracted into `currentContentPlainText`). PDF renders the attributed string through an `NSTextView` + `NSPrintOperation` with `jobDisposition = .save`, so pagination is handled by the print system. Word uses `NSAttributedString`'s `.officeOpenXML` document type.
- Dynamic system colors are resolved to their light-appearance values before writing, so exporting while the app is in dark mode doesn't yield white text on a white page. The app is activated before `runModal()` so the save dialog can't open behind other windows (the main window is a non-activating panel).
- `TranscriptFlow.attributed` (from #696) gains an optional `timestampColumnWidth` (default unchanged) because the export's `HH:mm:ss` needs a wider column than the on-screen `HH:mm`.

## Testing

- `swift build -c debug` clean
- Writer code exercised headlessly with an 80-line transcript: produces a valid 5-page PDF (verified page 1 renders with correct columns, colors, hanging indents) and a valid Word 2007+ `.docx` (content confirmed in `word/document.xml`)
- Manually exercised in the app: all three formats from each tab, save panel naming, empty-tab disabled state

### DIFF
--- a/OpenOats/Sources/OpenOats/Views/ContentExporter.swift
+++ b/OpenOats/Sources/OpenOats/Views/ContentExporter.swift
@@ -1,0 +1,156 @@
+import AppKit
+import SwiftUI
+import UniformTypeIdentifiers
+
+// MARK: - Export Format
+
+enum ExportFormat: CaseIterable {
+    case plainText
+    case pdf
+    case docx
+
+    var menuTitle: String {
+        switch self {
+        case .plainText: "Plain Text (.txt)"
+        case .pdf: "PDF (.pdf)"
+        case .docx: "Word (.docx)"
+        }
+    }
+
+    var fileExtension: String {
+        switch self {
+        case .plainText: "txt"
+        case .pdf: "pdf"
+        case .docx: "docx"
+        }
+    }
+
+    var contentType: UTType {
+        switch self {
+        case .plainText: .plainText
+        case .pdf: .pdf
+        case .docx: UTType("org.openxmlformats.wordprocessingml.document") ?? .data
+        }
+    }
+}
+
+// MARK: - Content Exporter
+
+/// Saves meeting content to a user-chosen file as plain text, PDF, or Word.
+@MainActor
+enum ContentExporter {
+    enum ExportError: LocalizedError {
+        case pdfGenerationFailed
+
+        var errorDescription: String? {
+            switch self {
+            case .pdfGenerationFailed: "The PDF could not be generated."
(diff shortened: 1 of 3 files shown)


yazinsai:
Thanks @Amir-Abushanab — this is a well-built addition, and a natural follow-up to #696.

What I checked and liked:
- No new dependency. PDF goes through `NSPrintOperation` so the print system handles pagination, and `.docx` rides `NSAttributedString`'s `.officeOpenXML` writer — both AppKit, both appropriate for a macOS-native app.
- Resolving dynamic system colors to their aqua values before writing is the kind of detail that usually gets missed; a dark-mode export would otherwise be white-on-white.
- `NSApp.activate(ignoringOtherApps:)` before `runModal()` is needed here because the main window is a non-activating panel. Good catch.
- Extracting `currentContentPlainText` so Copy and the `.txt` export share one string is the right factoring.
- `timestampColumnWidth` defaults to the existing 34, so #696's on-screen rendering is untouched.

One inconsistency worth a follow-up, not a blocker: `currentContentPlainText` uses `record.speaker.displayLabel` while `currentContentAttributedText` uses `displayName(speakerNames:)`. Renamed speakers therefore reach the PDF and Word exports but not the `.txt` one or the Copy button. That mirrors Copy's current behaviour, so it is defensible as-is — but #703 is fixing the same gap in the Markdown and plain-text repository exports, so this view path will end up the odd one out. Happy to take a small follow-up.

PR #700, opened 2026-08-27
Fix the "View" picker label wrapping to two lines in the detail toolbar

Amir-Abushanab:
## Bug

The "View" label next to the Transcript / Notes / Ask segmented control in the session detail toolbar wraps onto two lines, rendering as "Vie" / "w".

**Cause:** on macOS a SwiftUI `Picker` renders its title as visible text next to the control, and the `.frame(minWidth: 120, maxWidth: 220)` caps the *labeled* picker as a whole. The three segments consume essentially the full width, leaving the built-in label a sliver, so it wraps.

## Fix

Render the label as an explicit `Text("View")` with `.fixedSize()` beside the picker (so it can never wrap), and add `.labelsHidden()` to the picker — the title is kept for VoiceOver while the width cap now applies only to the segmented control.

### DIFF
--- a/OpenOats/Sources/OpenOats/Views/MeetingDetailPane.swift
+++ b/OpenOats/Sources/OpenOats/Views/MeetingDetailPane.swift
@@ -1510,12 +1510,18 @@ struct MeetingDetailPane<SessionFolderMenuItems: View>: View {
     private func detailToolbar(controller: NotesController, state: NotesState) -> some View {
         VStack(alignment: .leading, spacing: 8) {
             HStack(spacing: 8) {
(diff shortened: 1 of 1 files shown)


yazinsai:
Thanks @Amir-Abushanab — clean diagnosis. Rendering the title as an explicit `Text("View")` with `.fixedSize()` and moving `.labelsHidden()` onto the picker is exactly the right fix: the width cap now bounds only the segmented control, and the picker keeps its title for VoiceOver. Six lines, one file, no behaviour change beyond the layout. Merging.

PR #701, opened 2026-08-29
Prefer real meetings over overlapping personal blocks when binding a session to a calendar event

thunter009:
## The bug

`CalendarManager.currentEvent` ranks candidate events by how close their start is to the lookup instant, breaking ties on `startDate`:

```swift
.min { a, b in
    let distA = abs(a.startDate.timeIntervalSince(date))
    let distB = abs(b.startDate.timeIntervalSince(date))
    if distA != distB { return distA < distB }
    return a.startDate < b.startDate   // identical starts → always false
}
```

When two events start on the same clock boundary both keys tie, `min` keeps whichever element came first, and the answer is really EventKit's ordering. That is not a rare corner — personal blocks and meetings are both scheduled on the hour.

I hit it with a recording that started at 20:01:43:

| Event | Start | Attendees | Online |
|---|---|---|---|
| `Mimi - Prep Group Class` (Family Calendar) | 20:00:00 | 0 | no |
| the actual video call | 20:00:00 | 4 | yes |

The session bound to the family block. `session.json` came out with `"calendarEvent.title": "Mimi - Prep Group Class"`, `"participants": []`, `"isOnlineMeeting": false` — while the real meeting sat in the same candidate list carrying four invitees and a join link.

Note that overlap-based ranking would not have helped here either: the family block *contains* the recording start and covers 58 of its 75 minutes. Attendees and the conference link are the only signals that separate these two.

## The fix

Rank meeting-likeness — has invitees, or `isOnlineMeeting` — ahead of start proximity, then keep the existing proximity ordering, then prefer the shorter event (which the old doc comment claimed to do but never did), then `startDate` so the result is total and no longer depends on EventKit's ordering.

`CalendarEvent(from:)` already computes `participants` and `isOnlineMeeting`, so the discriminating signals cost nothing extra.

**Recording an in-person meeting still works the way it did.** Meeting-likeness only reorders candidates when at least one of them has it; with no meeting-like candidate, ranking falls straight back to closest start. The only outcomes that change are ones where a bare no-attendee block previously beat a real meeting.

The ranking moves into `CalendarEventMatcher`, a pure function over `CalendarEvent` values, so it is testable without an `EKEventStore`. The one behavioural cost is that all candidates are now converted to `CalendarEvent` before ranking rather than just the winner — that runs `NSDataDetector` over each event's notes. The lookup window is 30 minutes wide, so this is a handful of events at most.

## Testing

- `CalendarEventMatcherTests` — 7 new cases, including the tie above asserted in both input orderings, and the fallback that keeps no-attendee events working.
- Full suite: 708 tests, 0 failures.

## Possible follow-up, not included here

The event is resolved once at session start and passed through to `finalizeSession` unchanged, so a meeting that starts late binds against a point-in-time guess that the finished session's real interval could correct. That is a larger change and this fix stands on its own, so I left it out.

### DIFF
--- a/OpenOats/Sources/OpenOats/Meeting/CalendarManager.swift
+++ b/OpenOats/Sources/OpenOats/Meeting/CalendarManager.swift
@@ -55,7 +55,8 @@ final class CalendarManager {
 
     // MARK: - Event Lookup
 
-    /// Find the calendar event that best overlaps the given date (typically now).
+    /// Find the calendar event the given date (typically now) most likely belongs to.
+    /// Real meetings are preferred over personal blocks — see `CalendarEventMatcher`.
     /// Returns nil if no event is found or access is not authorized.
     func currentEvent(
         at date: Date = Date(),
@@ -74,20 +75,11 @@ final class CalendarManager {
             end: windowEnd,
             calendars: calendars
         )
-        let events = store.events(matching: predicate)
-
-        // Prefer the event whose start is closest to now, breaking ties by duration
-        let best = events
+        let candidates = store.events(matching: predicate)
             .filter { !$0.isAllDay }
-            .min { a, b in
-                let distA = abs(a.startDate.timeIntervalSince(date))
-                let distB = abs(b.startDate.timeIntervalSince(date))
-                if distA != distB { return distA < distB }
-                return a.startDate < b.startDate
-            }
+            .map { CalendarEvent(from: $0) }
 
-        guard let best else { return nil }
-        return CalendarEvent(from: best)
+        return CalendarEventMatcher.bestMatch(among: candidates, at: date)
     }
 
     /// Upcoming calendar events starting within the given time window, ordered by start date.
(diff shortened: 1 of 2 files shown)


yazinsai:
Thanks @thunter009 — this is a good find, and the write-up made it easy to verify.

The old comparator did claim to break ties by duration and then never did; `a.startDate < b.startDate` on two identical starts is always false, so `min` fell through to whatever order EventKit returned. Your `Mimi - Prep Group Class` case is the honest version of that bug: the family block and the video call both start on the hour, and the only signals separating them are invitees and the conference link — which, as you note, overlap-based ranking would not have fixed either, since the personal block actually contains more of the recording.

Two things I specifically checked:
- **In-person meetings still bind correctly.** Meeting-likeness is the first key, so it only reorders candidates when at least one is meeting-like. With no meeting-like candidate every event scores 1 and ranking falls straight through to start proximity, which is the old behaviour. The only outcomes that change are the ones that were wrong.
- **The ranking is now total.** The four-element tuple ends in `startDate`, so the result no longer depends on EventKit's ordering — and `CalendarEventMatcherTests` asserting the tie in *both* input orderings is the right way to prove that.

Moving the ranking into `CalendarEventMatcher` as a pure function over `CalendarEvent` is what makes this testable without an `EKEventStore`, and the extra `NSDataDetector` pass over a 30-minute window of candidates is a fair price.

Noted your follow-up about re-resolving the event at `finalizeSession` against the finished session's real interval. Agreed that it is a separate change and that this one stands on its own — please do open it if you want to take it.

PR #702, opened 2026-08-29
Stop splitting one speaker across "Them" and "Speaker 1"

thunter009:
## The bug

`DiarizationManager.dominantSpeaker` decides between `.them` and `.remote(n)` like this:

```swift
// If only one speaker was detected in the entire session, fall back to .them
// for backward compatibility (no point labeling "Speaker 1" when there's only one).
let activeSpeakers = speakers.values.filter { $0.hasSegments }
if activeSpeakers.count <= 1 {
    return .them
}
return .remote(bestSpeaker + 1)
```

The comment says "in the entire session", but this runs **live, once per utterance**, from `TranscriptionEngine`'s `onFinal` handler. `speakers` is the diarizer's running state, so `activeSpeakers.count` grows during the meeting and the branch flips partway through.

So one participant gets two labels. Everything they say before the diarizer registers a second cluster is stored as `them`; everything after is `remote_1`. Nothing goes back to fix the earlier rows.

From a two-person recording of mine, 820 utterances:

| Label | Count |
|---|---|
| `you` (me, mic) | 127 |
| `them` | 57 |
| `remote_1` | 636 |

Two people, three labels — `them` and `remote_1` are the same voice, just before and after the flip.

It also defeats the rename UI. `speakerNames` is keyed on `Speaker.storageKey`, so renaming `remote_1` to a real name leaves the 57 earlier rows still reading "Them", with no way to merge the two.

## The fix

Map the cluster index to a label directly, with no dependency on how many clusters exist *so far*:

```swift
nonisolated static func speaker(forDiarizerIndex index: Int) -> Speaker {
    index <= 0 ? .them : .remote(index + 1)
}
```

Cluster 0 stays `.them`, which is what the old fallback was protecting — a session with one remote voice still reads "Them", not "Speaker 1". Later clusters continue that numbering, so cluster 1 is "Speaker 2" (with "Them" as speaker 1).

`speakerRuns` carried a copy of the same fallback, hard-coded to `.them` regardless of which cluster was actually active. It now shares the mapping, so its single-speaker shortcut can no longer disagree with `dominantSpeaker`.

## Scope

Sessions already on disk keep the labels they were recorded with — there is no migration here, only a change to how new sessions are labelled. Happy to add one if you would rather old sessions were repaired too, though `Utterance` stores a wall-clock `timestamp` rather than an audio time range, so a retroactive re-query of the diarizer would need more than a rewrite pass.

The visible change for multi-speaker sessions is that the first remote cluster now reads "Them" instead of "Speaker 1". If you would rather keep "Speaker 1" and drop `.them` for multi-speaker recordings, that is a one-line change to the mapping and I am happy to switch it.

## Testing

- `DiarizerSpeakerMappingTests` — 4 new cases, including the invariant that the label depends only on the cluster index.
- Full suite: 705 tests, 0 failures.

### DIFF
--- a/OpenOats/Sources/OpenOats/Transcription/DiarizationManager.swift
+++ b/OpenOats/Sources/OpenOats/Transcription/DiarizationManager.swift
@@ -59,16 +59,23 @@ actor DiarizationManager {
 
         guard bestOverlap > 0 else { return .them }
 
-        // If only one speaker was detected in the entire session, fall back to .them
-        // for backward compatibility (no point labeling "Speaker 1" when there's only one).
-        let activeSpeakers = speakers.values.filter { $0.hasSegments }
-        if activeSpeakers.count <= 1 {
-            return .them
-        }
+        return Self.speaker(forDiarizerIndex: bestSpeaker)
+    }
 
-        // Map diarizer speaker index to Speaker enum
-        // Index 0 → .remote(1), index 1 → .remote(2), etc.
-        return .remote(bestSpeaker + 1)
+    /// Maps a diarizer cluster index to a stable `Speaker` label.
+    ///
+    /// The first cluster stays `.them`, so a session with one remote voice still
+    /// reads "Them" rather than "Speaker 1". Later clusters continue that numbering
+    /// — "Them" is speaker 1, so cluster 1 is "Speaker 2".
+    ///
+    /// The mapping deliberately depends only on the index, never on how many
+    /// clusters exist *so far*. `dominantSpeaker` runs live, once per utterance,
+    /// while the diarizer is still discovering speakers, so a "only one speaker
+    /// known yet, call it .them" rule labels one voice `.them` early and
+    /// `.remote(1)` from the moment a second voice appears. Nothing revisits the
+    /// earlier utterances, so a single participant ends up split across two labels.
+    nonisolated static func speaker(forDiarizerIndex index: Int) -> Speaker {
+        index <= 0 ? .them : .remote(index + 1)
     }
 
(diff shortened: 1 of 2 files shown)


yazinsai:
Thanks @thunter009 — the 820-utterance table makes the case better than any argument could. 127 / 57 / 636 across two people is exactly the signature of a branch flipping partway through a session.

The root cause is right: the comment said "in the entire session" but `dominantSpeaker` runs live from `TranscriptionEngine`'s `onFinal`, once per utterance, against the diarizer's *running* state. So `activeSpeakers.count <= 1` is a statement about what the diarizer knows so far, not about the session — and nothing revisits the utterances written before the second cluster appeared. The knock-on effect you describe on the rename UI is the part that makes this more than cosmetic: `speakerNames` is keyed on `storageKey`, so renaming `remote_1` strands the 57 earlier `them` rows with no way to merge them.

Making the mapping a pure function of the cluster index is the correct fix, and keeping cluster 0 as `.them` preserves what the old fallback was actually protecting — a one-remote-voice session still reads "Them", not "Speaker 1". Folding `speakerRuns`' duplicate hard-coded `.them` into the same mapping removes the possibility of the two disagreeing, which was a real latent bug of its own.

On the two open questions you raised:
- **"Them" vs "Speaker 1" for the first cluster in multi-speaker sessions:** keeping `.them` is the right call. It is the least-surprising behaviour for the common two-person case, and it is what existing sessions already read.
- **Migration for sessions already on disk:** agreed, leave it out. As you say, `Utterance` stores a wall-clock timestamp rather than an audio time range, so a retroactive re-query is not a rewrite pass. New sessions being correct is the win here.

PR #703, opened 2026-08-29
Use speaker names in exports, and seed them from the calendar event

thunter009:
Two commits, split so the first can be taken on its own.

## 1. Renamed speakers never reached exports

`SessionIndex.speakerNames` already exists, and the transcript UI already writes to it — clicking a speaker label opens a rename field that calls `updateSessionSpeakerNames`. Nothing downstream ever read it.

Both export paths rendered `Speaker.displayLabel` directly:

```swift
// MarkdownMeetingWriter
static func speakerLabel(_ speaker: Speaker) -> String { speaker.displayLabel }

// SessionRepository.exportPlainText
result += "[\(timeFmt.string(from: record.timestamp))] \(record.speaker.displayLabel): \(displayText)\n"
```

So a session where every speaker had been named by hand still exported as "You" and "Them" — in the frontmatter `participants` list and on every transcript line. The rename only ever showed up in the app.

`Speaker.displayName(speakerNames:)` was already written for exactly this and was unused outside the detail pane. This threads the session's names through `MarkdownMeetingWriter.Metadata` and into `exportPlainText`, and calls it. The mirror path built its `SessionIndex` field by field and dropped `speakerNames` on the way, so that is passed through too.

Sessions with no names are unchanged — `displayName` falls back to `displayLabel`.

This is what `docs/meeting-format-spec.md` calls the "Future State: Named Participants"; the writer already supported it structurally.

## 2. Seed the names from the calendar event

Nothing populated `speakerNames` automatically. When a session is bound to a calendar event the invitee list is already in `CalendarEvent.participants`, and `invitedParticipantDisplayNames` already resolves it to display names.

Seeding is deliberately narrow. Names are assigned only when the mapping is **forced**: one remote voice in the transcript, and one invitee left after removing the recorder (`NSFullUserName()`, which the writer already uses for the `recorder:` field). Two voices and three invitees carry no signal about which diarizer cluster belongs to which person, so that case is left alone — a confidently wrong name on a transcript is worse than a generic one.

Names the user set by hand always win; seeding only ever fills an empty field.

In practice this covers 1:1 calls, which is where the naming matters most and where the answer is unambiguous.

## Testing

- `SpeakerNameSeederTests` — 7 cases, mostly about *declining* to guess.
- `MarkdownMeetingWriterSpeakerNameTests` — 4 cases covering transcript lines and the frontmatter participants list, with and without names.
- Full suite: 712 tests, 0 failures.

## Note on ordering

This is independent of #701 and #702 and can be merged in any order. It is most useful with #701, since seeding reads whichever calendar event the session bound to — if that binding is wrong, the seeded name is wrong too. The narrow one-voice/one-invitee rule limits the blast radius, but #701 is what makes the input trustworthy.

### DIFF
--- a/OpenOats/Sources/OpenOats/Intelligence/MarkdownMeetingWriter.swift
+++ b/OpenOats/Sources/OpenOats/Intelligence/MarkdownMeetingWriter.swift
@@ -25,6 +25,8 @@ enum MarkdownMeetingWriter {
         let language: String?
         let meetingApp: String?
         let engine: String?
+        /// Per-session speaker display names, keyed by `Speaker.storageKey`.
+        let speakerNames: [String: String]?
 
         init(from index: SessionIndex) {
             self.sessionID = index.id
@@ -34,6 +36,7 @@ enum MarkdownMeetingWriter {
             self.language = index.language
             self.meetingApp = index.meetingApp
             self.engine = index.engine
+            self.speakerNames = index.speakerNames
         }
     }
 
@@ -100,7 +103,13 @@ enum MarkdownMeetingWriter {
     static func buildMarkdown(metadata: Metadata, records: [SessionRecord], notesMarkdown: String? = nil) -> String {
         let resolvedTitle = metadata.title?.isEmpty == false ? metadata.title! : "Meeting"
         let frontmatter = buildFrontmatter(metadata: metadata, records: records, title: resolvedTitle)
-        let body = buildBody(title: resolvedTitle, records: records, startedAt: metadata.startedAt, notesMarkdown: notesMarkdown)
+        let body = buildBody(
+            title: resolvedTitle,
+            records: records,
+            startedAt: metadata.startedAt,
+            notesMarkdown: notesMarkdown,
+            speakerNames: metadata.speakerNames
+        )
         return frontmatter + "\n" + body
     }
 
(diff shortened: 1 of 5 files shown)


yazinsai:
Thanks @thunter009 — both halves of this are well argued, and the second one is careful in exactly the right place.

On the first commit: you are right that `speakerNames` was written by the rename UI and then read by nobody outside the detail pane. `Speaker.displayName(speakerNames:)` already existed for this and threading it through `MarkdownMeetingWriter.Metadata` and `exportPlainText` is the minimal fix. Catching that the mirror path rebuilt `SessionIndex` field by field and dropped `speakerNames` on the way is the detail that would have made this half-work otherwise.

On the seeding: the restraint is the reason I am comfortable taking this autonomously. Assigning a name only when the mapping is *forced* — one remote voice, one invitee after removing the recorder — means the feature never guesses, and `SpeakerNameSeederTests` being mostly cases about declining to guess is the right shape for a test suite here. Hand-set names winning over seeded ones is also correct.

I checked the one thing that would have made this `risk:high`: whether a real invitee name starts reaching cloud providers. It does not. `NotesEngine`, `SuggestionEngine`, `BatchTextCleaner`, and `SidecastEngine` all still build their prompts from `speaker.displayLabel`, so seeded names stay in local session storage, the local Markdown mirror, and the plain-text export. That keeps this inside the local-first guarantee.

PR #704, opened 2026-08-30
Fix total notes failure when a reasoning model returns null content

thunter009:
## Root cause

`CompletionResponse.CompletionMessage.content` is declared non-optional (`OpenOats/Sources/OpenOats/Intelligence/OpenRouterClient.swift:452`). A reasoning model served via an OpenAI-compatible endpoint (e.g. Qwen3 on `mlx_lm.server` with thinking enabled) returns `"content": null` with its output in a sibling `"reasoning"` field. `JSONDecoder` then throws at the decode site (`:273`), which surfaces as a total notes-generation failure. The streaming path is unaffected — `SSEChunk.Delta.content` (`:440`) is already `String?`.

## What changed

- `CompletionMessage.content` is now `String?`, and `reasoning: String?` is decoded alongside it.
- The decode + fallback logic moved into an internal `static completionText(from:)` so it is unit-testable, same shape as the existing `preflightError(for:apiKey:)` static.
- Normal fallback is `content ?? ""`. When content is nil/empty but `reasoning` is non-empty, a new `OpenRouterError.reasoningOnlyResponse` is thrown whose description names the likely cause (thinking mode enabled server-side) instead of silently producing empty notes.

## Testing

- 5 new decoder tests in `OpenRouterClientTests` with JSON fixtures: normal content; content + reasoning both present (content wins); `"content": null` + reasoning (throws `reasoningOnlyResponse`, description checked); empty-string content + reasoning (throws); null content with no reasoning (returns `""`, matching the old `?? ""` intent).
- `swift build -c debug`, `swift test` (728 tests, 0 failures), and `swift build -c release` all pass locally (macOS, arm64).

## Risk

Low, confined to the non-streaming completion path. Well-formed responses decode identically. Responses that previously crashed the decoder now either recover (content present) or fail with an actionable error message.

---

## Revised after self-review

Re-reading the call graph showed the original patch did not fix the reported bug. `completionText` is only reachable from `complete()` (`SidecastEngine:86`, `SuggestionEngine:176`, `LiveTranscriptCleaner:198`, `BatchTextCleaner:292`, `MeetingDetailPane:3873`). Notes generation does not go through any of them: `NotesEngine.generate` calls `streamCompletion` (`NotesEngine.swift:222`). The "streaming path is unaffected" line in the root-cause section above is wrong — `SSEChunk.Delta.content` being optional is exactly why the streaming failure was silent rather than loud. This revision fixes the streaming path as well.

**Streaming path.** `SSEChunk.Delta` now decodes the sibling `reasoning` field. `streamCompletion` records whether the stream ever delivered a content delta and whether it ever delivered a reasoning delta. A stream that delivered reasoning but no content finishes with `reasoningOnlyResponse` instead of finishing normally and handing the caller an empty string.

**Empty notes are a failure.** `NotesEngine` now rejects a completed stream whose accumulated markdown is empty or whitespace-only. This is deliberately generic: it catches a reasoning-only stream, a token budget exhausted before any notes were written, and any other shape of "the model produced nothing", without needing to recognise the provider's flavour of it.

**Trim asymmetry.** The first pass trimmed `reasoning` before testing it but tested `content` untrimmed, so a response of `"content": "\n"` beside a full reasoning block slipped through as a successful empty answer. Emptiness is now tested on the trimmed text; non-blank content is still returned untrimmed.

**No more silent empty string.** Before this PR, `"content": null` with no `reasoning` threw a `DecodingError`, which the Ask pane surfaced as a visible `askError`. Making `content` optional turned that into a silent empty bubble. Fully-empty responses now throw `emptyResponse` ("Empty response from server — the model returned no content"), so that regression is gone and empty `choices` is covered too.

**Error copy.** The cleaners call `complete(maxTokens: 512)`, so hitting the token budget mid-reasoning is at least as likely as server-side thinking mode. The message no longer asserts one cause: "The response contained only reasoning output and no final answer. Thinking mode may be enabled on the server, or the token budget was exhausted before the answer was written."

**Where this copy is seen.** Today it reaches users directly only through the Ask pane, which renders `askError`. Everything else logs it to diagnostics. If #707 also lands, it additionally reaches the notes-failure notification, which carries the underlying `localizedDescription` — and in that auto-notes path #707's notification title fronts this message.

**Testing.** 737 tests, 0 failures (was 728). New coverage: whitespace-only content beside reasoning, untrimmed passthrough of non-blank content, `emptyResponse` for null content with no reasoning and for empty `choices`, the reworded description, the stream outcome tracker (reasoning-only, content-arrives, ordinary stream, empty deltas), and `generateMarkdownDetached` failing on whitespace-only output.

## Merge order

Merge this before #707. #707 relies on the `reasoningOnlyResponse` error introduced here to classify a
reasoning-only completion as non-retriable; without this branch it falls back to treating that case as a
generic failure. No other open PR touches these files.

### DIFF
--- a/OpenOats/Sources/OpenOats/Intelligence/NotesEngine.swift
+++ b/OpenOats/Sources/OpenOats/Intelligence/NotesEngine.swift
@@ -277,6 +277,14 @@ final class NotesEngine {
                 }
                 if let error = self.error {
                     continuation.resume(throwing: GenerationError.failed(error))
+                } else if self.generatedMarkdown.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty {
+                    // The stream finished without producing notes. Thinking mode and an
+                    // exhausted token budget both land here, and there is nothing worth
+                    // saving either way, so fail instead of storing an empty note.
+                    continuation.resume(throwing: GenerationError.failed(
+                        "The model returned empty notes. Thinking mode may be enabled on the "
+                        + "server, or the token budget was exhausted before the notes were written."
+                    ))
                 } else {
                     continuation.resume(returning: self.generatedMarkdown)
                 }
--- a/OpenOats/Sources/OpenOats/Intelligence/OpenRouterClient.swift
+++ b/OpenOats/Sources/OpenOats/Intelligence/OpenRouterClient.swift
@@ -270,8 +270,22 @@ actor OpenRouterClient {
             throw OpenRouterError.httpError(statusCode, host: targetURL.host)
         }
 
+        return try Self.completionText(from: data)
+    }
+
+    /// Decodes a non-streaming chat completion, tolerating reasoning models
+    /// (e.g. Qwen3 served by mlx_lm.server) that return `"content": null` with
+    /// their output in a sibling `reasoning` field instead.
+    static func completionText(from data: Data) throws -> String {
         let completionResponse = try JSONDecoder().decode(CompletionResponse.self, from: data)
-        return completionResponse.choices.first?.message.content ?? ""
+        let message = completionResponse.choices.first?.message
+        let content = message?.content ?? ""
+        if content.isEmpty,
+           let reasoning = message?.reasoning,
+           !reasoning.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty {
+            throw OpenRouterError.reasoningOnlyResponse
+        }
+        return content
     }
 
     private func streamAnthropicCompletion(
@@ -404,6 +418,7 @@ actor OpenRouterClient {
     enum OpenRouterError: Error, LocalizedError {
         case httpError(Int, host: String?)
         case missingAPIKey(host: String?)
+        case reasoningOnlyResponse
 
         var errorDescription: String? {
             switch self {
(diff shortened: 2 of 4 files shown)


yazins-ai:

The most valuable thing in this PR is the retraction. The first version fixed `completionText`, and your own re-reading of the call graph found that notes generation never goes through it — `NotesEngine.generate` calls `streamCompletion`. So the original patch would have shipped, closed the issue, and left the reported bug exactly where it was. Catching that yourself, and saying plainly that the "streaming path is unaffected" line was wrong, is worth more than the fix.

The second insight follows from it and is the sharper one: the streaming path being optional already is *why* the failure was silent rather than loud. A non-optional field throws and you get a stack trace; an optional field quietly yields empty notes and the user is left guessing. Turning an empty accumulated stream into a failure closes that class, not just this instance.

`reasoningOnlyResponse` naming thinking-mode as the likely cause is the right level of detail for an error a user has to act on — it points at a server-side setting they can actually change.

The five decoder tests with real JSON fixtures, including null-content-with-no-reasoning still returning empty to match the old `?? ""` intent, made the behavior change easy to check against the old contract.

PR #705, opened 2026-08-30
Fix retained batch audio growing without bound (~23 MB/min): file-based TTL, periodic sweep, retention setting

thunter009:
## Problem

When `enableBatchRetranscription` is on, `stashAudioForBatch` moves `mic.caf`/`sys.caf` into `<session>/audio/` permanently (`Storage/SessionRepository.swift:1387`, moves at `:1397`/`:1401`). Retained audio is 48kHz Float32 mono, two stems — up to about **23 MB per meeting minute** in theory; one real 75-minute session measured **1.0 GB** on disk (232 MB mic + 869 MB sys). A 7-day TTL sweep exists (`cleanupExpiredRetainedBatchAudio` at `SessionRepository.swift:2116`, constant at `:180`), but it has two defects that let audio grow without bound in practice:

1. **Wrong mtime source.** `SessionRepository.swift:2126-2127` reads `contentModificationDateKey` off the *session directory*, not the audio files. Any metadata write — a notes edit, a speaker rename — bumps the directory mtime and resets the TTL. A session you keep touching never sheds its audio.
2. **Launch-only.** The sweep's single call site is the repository initializer (`SessionRepository.swift:230`). A menu-bar app that stays running for weeks never prunes.

Additionally, `cleanupBatchAudio` (`SessionRepository.swift:1467`) — the per-session removal helper — had zero call sites, and there was no user-facing retention control (`Views/SettingsView.swift:580-595` only had the batch toggle and model picker).

## Changes

- **Expiry judged by the retained files themselves.** `newestRetainedBatchAudioModificationDate(inSessionDirectory:)` takes the newest mtime among `mic.caf`/`sys.caf`/`batch-meta.json` (canonical and legacy layouts); the session directory's date is deliberately ignored. The sweep deletes only when that newest artifact is past the retention window.
- **Periodic sweep.** The init-time sweep stays; the repository also runs the sweep every 6 hours via a repeating `Task` owned by the actor, cancelled in `deinit`. Changing the retention setting also triggers an immediate sweep.
- **User-configurable retention.** New `RetainedBatchAudioRetention` setting (3/7/14/30 days or *Keep forever*), default **7 days** — current behavior preserved. Stored in UserDefaults following the existing `SettingsStore` pattern; surfaced next to the batch toggle in Settings → Transcript Quality. The sweep re-reads the value each pass, so changes apply without a relaunch.
- **Per-session "Delete retained audio".** First call site for `cleanupBatchAudio`: a destructive action (with confirmation) in the transcript maintenance menu next to the re-transcribe affordance in `Views/MeetingDetailPane.swift`. It stops playback of a stem before unlinking it, refreshes playback sources, clears the post-meeting "Re-transcribe" banner offer, and flips `canRetranscribeSelectedSession` off (the `hasRetainedBatchAudio` gate at `App/NotesController.swift:347`).

## Revised after self-review

An adversarial pass over the first version found fail-open paths in a destructive feature. Principle applied throughout: **unknown state must mean keep, never delete.**

- **Fail-closed default.** `SessionRepository.init`'s `batchAudioRetention` closure now defaults to `{ nil }` (keep forever). Previously it defaulted to the 7-day lifetime, so every constructor path that did not wire the setting (the plain test/UI-test constructors, any future call site) silently became an unconditional deleter. Only the wired `AppContainer` path supplies a real retention, threaded from the actual `SettingsStore` instance (not a hardcoded `UserDefaults.standard`).
- **Fail-closed decode.** An unknown raw value (downgrade path) or a non-string in UserDefaults previously decoded to the 7-day default — turning a user's "Keep forever" into deletion. Unknown/corrupt now decodes to `.forever`; only a genuinely-unset key gets the 7-day default (`RetainedBatchAudioRetention.stored(in:)` and the `SettingsStore` accessor twin).
- **Sweep vs. in-flight batch run.** `BatchAudioTranscriber` captures stem URLs once at the start of a run and re-opens `sys.caf` minutes later; a sweep tick mid-run would make the batch fail unretryably with the audio already gone. Rather than a cross-actor status read, `process()` now brackets the run with `beginBatchAudioAccess`/`endBatchAudioAccess` on the repository actor itself (counted, so overlapping runs for the same session keep the guard up). The sweep skips guarded sessions and `cleanupBatchAudio` refuses (returns `false`) while a run holds the stems. This is covered by an actor-level test; no cross-actor read is needed because the repository is already the party both sides talk to.
- **Action-time re-gating.** The delete menu item's `.disabled(isBatchBusy)` is render-time only; `deleteRetainedBatchAudio()` now re-checks `coordinator.batchStatus == .idle` when invoked, and the repository-side refusal above backstops the remaining race.
- **State consistency.** Deleting audio clears `lastEndedSessionCanRetranscribe`, so the post-meeting banner cannot offer a re-transcribe that would no-op. Periodic-sweep removals notify the UI (repository callback → `.retainedBatchAudioSwept` notification), and open detail panes re-derive audio sources and the re-transcribe gate — the same code path as the manual delete.
- **Playback ordering.** Playback of a doomed stem stops *before* unlinking; the user's selected audio source is kept when it survives the deletion instead of being reset.
- **Symlink guard.** The sweep skips symlinks: `resourceValues` follows links, so a symlink named `session_*` would have deleted its *target's* stems.
- **Orphan metadata.** A lone `batch-meta.json` with no stems used to survive forever (no reference date); it is now removed once its own mtime passes the window.
- **Honest size math.** The settings copy and this PR now say "up to about 23 MB per meeting minute; one real 75-minute meeting measured 1.0 GB" — the earlier "roughly 1 GB" phrasing implied 23×75, which is ~1.7 GB theoretical, not the measured figure.

**Considered and rejected: grandfathering pre-upgrade sessions to "Keep forever" on first launch.** The 7-day deletion at launch is the already-shipped default contract; the directory-mtime bug merely exempted recently-touched sessions from it. This PR makes the existing contract hold uniformly rather than introducing a new one. If you'd prefer a migration that pins existing sessions' audio, happy to add it.

## Tests

- `Tests/OpenOatsTests/RetainedBatchAudioSweepTests.swift` (14 tests): reference-date selection (audio-file mtime wins over a freshly-bumped directory; newest file wins; nil without stems), sweep regression for the directory-mtime defect and its inverse, keep-forever, legacy layout, non-session directories, lone expired/fresh `batch-meta.json`, active-batch exclusion, symlink skip, and the actor-level begin/end guard (sweep and user delete both refuse until the run ends).
- `SettingsStoreTests`: default, round-trip, and decode fallbacks (unset → 7 days; garbage string → forever; non-string → forever).
- `SessionRepositoryTests`: init-sweep tests now inject the retention explicitly; a new test proves the unwired constructor path never deletes.

`swift build -c debug`, `swift test` (**745 tests, 0 failures**), and `swift build -c release` all pass.

## Second review pass

An independent review of the revised branch raised two more issues in this diff. Both were real; both are fixed in `a24e97e`.

- **Re-check the selected session after each await.** `refreshRetainedAudioDependentState` (`App/NotesController.swift`) validated `state.selectedSessionID` once, before its two repository awaits. Either await yields the main actor, so a selection change while the sweep refresh is suspended applied the swept session's audio sources, audio URL and re-transcribe availability to whatever was selected on resume. The refresh now re-checks after each hop and bails rather than writing. Covered by `testRetentionSweepRefreshDoesNotApplyToASessionDeselectedMidRefresh`, which fails on the previous commit with all three leak assertions.

- **The retention setting was a real data race, not a stale read.** The sweep runs on the `SessionRepository` actor and reads the window through the `nonisolated` `retainedBatchAudioRetentionInterval`, which read `nonisolated(unsafe) private var _retainedBatchAudioRetention: String` while the main-actor picker setter wrote the same storage. ThreadSanitizer confirms it — `data race SettingsStore.swift:807` (setter) against `SettingsStore.swift:816` (getter), with the test process aborting on the report. Changing the picker during a sweep is enough to hit it. The backing storage is now an `OSAllocatedUnfairLock<String>`, the same primitive already used in `MicCapture` and `SystemAudioCapture`, so the escape hatch is removed rather than the window narrowed. Both accessors go through the lock, so there is still one source of truth. TSan is clean afterwards.

`handleRetainedAudioSwept` lost its `private` so the sweep path can be driven directly from a test.

## Merge order

Merge before #710. Both branches edit `SessionRepository.swift` and conflict there (3 hunks plus tests):
this PR adds the audio-retention sweep, #710 adds the mirror-retry ledger. Order is not important to the
behaviour, only to the rebase — taking this one first means #710 rebases onto the retention changes.

### DIFF
--- a/OpenOats/Sources/OpenOats/App/AppContainer.swift
+++ b/OpenOats/Sources/OpenOats/App/AppContainer.swift
@@ -71,13 +71,27 @@ final class AppContainer {
                     .appendingPathComponent("Documents/OpenOats", isDirectory: true)
             )
             let settings = AppSettings()
-            let coordinator = AppCoordinator(
-                sessionRepository: SessionRepository(
-                    batchAudioRetention: {
-                        RetainedBatchAudioRetention.stored(in: .standard).timeInterval
-                    }
-                )
+            let sessionRepository = SessionRepository(
+                batchAudioRetention: { settings.retainedBatchAudioRetentionInterval }
             )
+            let coordinator = AppCoordinator(sessionRepository: sessionRepository)
+            Task {
+                // When the periodic sweep removes retained audio, re-derive the
+                // UI state that depends on it (post-meeting banner, detail pane).
+                await sessionRepository.setRetainedAudioSweepHandler { sessionIDs in
+                    Task { @MainActor in
+                        for sessionID in sessionIDs {
+                            coordinator.liveSessionController?
+                                .retainedBatchAudioWasDeleted(sessionID: sessionID)
+                        }
+                        NotificationCenter.default.post(
+                            name: .retainedBatchAudioSwept,
+                            object: nil,
+                            userInfo: ["sessionIDs": sessionIDs]
+                        )
+                    }
+                }
+            }
             let updaterController = AppUpdaterController()
             return AppLaunchContext(
                 isFirstLaunch: false,
--- a/OpenOats/Sources/OpenOats/App/AppContainer.swift
+++ b/OpenOats/Sources/OpenOats/App/AppContainer.swift
@@ -71,7 +71,13 @@ final class AppContainer {
                     .appendingPathComponent("Documents/OpenOats", isDirectory: true)
             )
             let settings = AppSettings()
-            let coordinator = AppCoordinator()
+            let coordinator = AppCoordinator(
+                sessionRepository: SessionRepository(
+                    batchAudioRetention: {
+                        RetainedBatchAudioRetention.stored(in: .standard).timeInterval
+                    }
+                )
+            )
             let updaterController = AppUpdaterController()
             return AppLaunchContext(
                 isFirstLaunch: false,
--- a/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
+++ b/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
@@ -1595,6 +1595,13 @@ final class LiveSessionController {
     /// Assigns `value` to `state[keyPath:]` only when it differs, avoiding spurious
     /// @Observable withMutation notifications that would trigger unnecessary layout passes.
     @inline(__always)
+    /// Clears the post-meeting "Re-transcribe" offer when the retained audio
+    /// backing it is deleted (user action or retention sweep).
+    func retainedBatchAudioWasDeleted(sessionID: String) {
+        guard coordinator.lastEndedSession?.id == sessionID else { return }
+        set(\.lastEndedSessionCanRetranscribe, false)
+    }
+
     private func set<T: Equatable>(_ kp: ReferenceWritableKeyPath<LiveSessionState, T>, _ value: T) {
         if state[keyPath: kp] != value { state[keyPath: kp] = value }
     }
--- a/OpenOats/Sources/OpenOats/App/NotesController.swift
+++ b/OpenOats/Sources/OpenOats/App/NotesController.swift
@@ -942,6 +942,12 @@ final class NotesController {
         guard state.selectedSessionID == sessionID else { return }
 
         let sources = await coordinator.sessionRepository.audioSources(for: sessionID)
+        // Every await here yields the main actor, and the sweep fires while the
+        // user is working, so the selection can move on before we resume.
+        // Applying this session's audio state to whatever is selected now is
+        // worse than skipping the refresh, so re-check after each hop.
+        guard state.selectedSessionID == sessionID else { return }
+
         if let current = state.audioFileURL, !sources.contains(where: { $0.url == current }) {
             stopAudio()
             state.audioFileURL = sources.first?.url
@@ -949,11 +955,15 @@ final class NotesController {
             state.audioFileURL = sources.first?.url
         }
         state.availableAudioSources = sources
-        state.canRetranscribeSelectedSession =
-            await coordinator.sessionRepository.hasRetainedBatchAudio(sessionID: sessionID)
+
+        let canRetranscribe = await coordinator.sessionRepository.hasRetainedBatchAudio(sessionID: sessionID)
+        guard state.selectedSessionID == sessionID else { return }
+        state.canRetranscribeSelectedSession = canRetranscribe
     }
 
-    private func handleRetainedAudioSwept(sessionIDs: [String]) {
+    /// Entry point for the `.retainedBatchAudioSwept` notification.
+    /// Not private so tests can drive the sweep path directly.
+    func handleRetainedAudioSwept(sessionIDs: [String]) {
         guard let sessionID = state.selectedSessionID, sessionIDs.contains(sessionID) else { return }
         Task { await refreshRetainedAudioDependentState(sessionID: sessionID) }
     }
--- a/OpenOats/Sources/OpenOats/App/NotesController.swift
+++ b/OpenOats/Sources/OpenOats/App/NotesController.swift
@@ -897,6 +897,26 @@ final class NotesController {
         }
     }
 
+    func deleteRetainedBatchAudio() {
+        guard let sessionID = state.selectedSessionID,
+              state.canRetranscribeSelectedSession else { return }
+
+        Task {
+            await coordinator.sessionRepository.cleanupBatchAudio(sessionID: sessionID)
+            guard state.selectedSessionID == sessionID else { return }
+
+            // The deleted stems may back the playback sources currently shown.
+            let sources = await coordinator.sessionRepository.audioSources(for: sessionID)
+            if let playingURL = state.audioFileURL, !sources.contains(where: { $0.url == playingURL }) {
+                stopAudio()
+            }
+            state.availableAudioSources = sources
+            state.audioFileURL = sources.first?.url
+            state.canRetranscribeSelectedSession =
+                await coordinator.sessionRepository.hasRetainedBatchAudio(sessionID: sessionID)
+        }
+    }
+
     // MARK: - Session Management
 
     func renameSession(sessionID: String, newTitle: String) {
--- a/OpenOats/Sources/OpenOats/App/NotesController.swift
+++ b/OpenOats/Sources/OpenOats/App/NotesController.swift
@@ -188,16 +188,32 @@ final class NotesController {
     @ObservationIgnored private var audioPlayer: AVPlayer?
     @ObservationIgnored private var playerObservation: Any?
 
+    /// Observer for retention-sweep removals, so open UI re-derives its state.
+    @ObservationIgnored nonisolated(unsafe) private var sweepObservation: (any NSObjectProtocol)?
+
     init(coordinator: AppCoordinator, settings: AppSettings? = nil) {
         self.coordinator = coordinator
         self.settings = settings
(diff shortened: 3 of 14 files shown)


yazins-ai:
Thanks @thunter009 — the diagnosis here is solid, and the two defects you identified are real: the directory-mtime TTL reset and the launch-only sweep together do let retained audio grow without bound. The 1.0 GB measurement on a 75-minute session makes the case well.

Labeled `risk:high` and `state:awaiting-human`. Two reasons, both about blast radius rather than the idea:

1. **Scope.** 14 files across `Storage`, `Settings`, `Views`, `App`, and `Transcription`, adding a new user-facing setting and a destructive per-session action. Repo policy routes that to a maintainer regardless of how good the change is.
2. **Two fail-open paths in the deletion logic**, which matter more than usual because this feature unlinks user audio on a timer. Your own stated principle is "unknown state must mean keep, never delete", and these two spots do not hold it:

**(a) An unreadable mtime is indistinguishable from a missing file.**

`modificationDate(of:)` returns `nil` both when the file is absent and when `resourceValues` throws. In `retainedBatchAudioDates`, if `mic.caf` and `sys.caf` are both present but their mtime reads fail, `stems` comes back `nil`. The sweep then falls into the `else if let metadata` branch, and an expired `batch-meta.json` deletes stems whose age was never established:

```swift
if let stems = dates.stems {
    let newest = max(stems, dates.metadata ?? stems)
    guard newest < cutoff else { continue }
} else if let metadata = dates.metadata {
    // No stems left: a lone batch-meta.json would otherwise live forever.
    guard metadata < cutoff else { continue }
}
```

The comment says "no stems left", but the branch is also reached when stems exist and are simply unreadable. Distinguishing "absent" from "unreadable" — e.g. having the helper return an enum, and treating `unreadable` as keep — closes it.

**(b) The symlink guard does not cover `audio/`.**

The guard checks the `session_*` entry itself, but `removeRetainedBatchAudio` then resolves `<session>/audio/mic.caf`. If `audio/` is a symlink, `removeItem` deletes through it. That is not a contrived setup for this feature specifically: a user carrying 1 GB per meeting is exactly the person likely to symlink `audio/` onto another volume. The same `isSymbolicLink` check applied to `audioDir` before the canonical-layout removals would cover it.

Neither is a reason to drop the change — the underlying bug is worth fixing and the ledger/backoff work in the sibling PRs shows the same care. But a maintainer should make the call on a feature that deletes recordings, so this one waits for @yazinsai. Fixing (a) and (b) in the meantime would make that review shorter.

thunter009:
Both fail-open paths are fixed, and the branch is rebased on current `main` (it was conflicting, it is not now).

**(a) Unreadable mtime no longer reads as absent.**

`modificationDate(of:)` returns an enum instead of an Optional:

```swift
enum ModificationDateResult: Equatable {
    case date(Date)
    case absent
    case unreadable
}
```

`absent` is decided by `CocoaError.fileReadNoSuchFile` / `.fileNoSuchFile`. Every other throw, and a successful read with no date attached, is `unreadable`. `RetainedBatchAudioDates` now carries `hasUnreadable`, and expiry refuses outright when it is set:

```swift
static func retainedBatchAudioIsExpired(_ dates: RetainedBatchAudioDates, cutoff: Date) -> Bool {
    guard !dates.hasUnreadable else { return false }
    ...
}
```

The `else if let metadata` branch you quoted is now only reachable when the stems are genuinely absent, which is what its comment always claimed.

**(b) The symlink guard covers `audio/`.**

`removeRetainedBatchAudio` checks `audio/` before any removal and refuses the session rather than redirecting it:

```swift
let audioDir = sessionDirectory.appendingPathComponent("audio", isDirectory: true)
if isSymbolicLink(audioDir) { return false }
```

It returns `Bool` now, so a refusal propagates. The sweep does not count a refused session as swept, and the per-session delete returns the refusal to its caller instead of reporting success.

**One thing worth flagging about the tests.**

My first test for (a) was vacuous and I nearly shipped it. It staged an unreadable mtime with `chmod 000`, but the same permission bits that hide the mtime also block `removeItem`, so the test passed with the guard deleted. The filesystem was doing the work the guard was supposed to do.

The fix was to make the decision testable directly: `retainedBatchAudioIsExpired` is pure, and the three fail-open cases are asserted against it rather than through a filesystem state that cannot be staged without also preventing deletion. I then re-checked every new guard by deleting it and confirming the suite goes red. The end-to-end symlink and unreadable-directory tests are still there, and I have noted in the file which of them are the weaker signal.

`swift test` — 828 tests, 0 failures.

Your scope point stands unchanged and I am not asking around it. Fourteen files, a new user-facing setting, and an action that deletes recordings on a timer is a maintainer call. This just makes that review shorter.

yazins-ai:
Re-triaged under the updated repo policy. Staying `risk:high` / `state:awaiting-human`, but **both reasons in my original brief are now void**, and a maintainer should not review this against stale reasoning.

**Reason 1 (scope) no longer exists as a policy matter.** I wrote that "14 files across 5 subsystems routes this to a maintainer regardless of how good the change is." That is no longer true: file count across subsystems is now explicitly "at least `risk:medium`, but not `risk:high` unless it meets one of the criteria above." The count on its own routes nothing.

**Reason 2 (the two fail-open paths) is fixed.** I verified both in the diff rather than taking the summary:

- `ModificationDateResult` distinguishes `absent` from `unreadable` via `CocoaError.fileReadNoSuchFile`, and `retainedBatchAudioIsExpired` refuses outright on `guard !dates.hasUnreadable`. The `else if let metadata` branch is now only reachable when stems are genuinely gone, which is what its comment always claimed.
- `removeRetainedBatchAudio` has `if isSymbolicLink(audioDir) { return false }` before any removal — refusing the session rather than redirecting it.

All three checks green, mergeable, no conflicts.

**What it still rests on, and it is one thing:** this feature unlinks user recordings on a timer, and deleted audio is not recoverable by a revert or a label flip. That is the category the escalation boundary exists for, so I am not merging it autonomously.

One correction that should make the call easy: **this does not change the retention default.** `main` already declares `retainedBatchAudioLifetime = 7 * 24 * 3600`, and the picker defaults to `.sevenDays`. The 7-day policy is pre-existing and documented; the bug is that the directory-mtime reset and launch-only sweep meant it was never actually enforced. So the change makes the stated default real rather than introducing a new one.

The practical consequence worth weighing: users whose audio survived indefinitely *because of the bug* will have it deleted at 7 days once this ships, as was always intended. That is the decision.

**Options**

1. **Merge as-is.** Enforces the documented 7-day default, setting available to extend to 30 days. Users relying on the bug lose audio at 7 days with no warning.
2. **Merge with a one-time notice** before the first sweep deletes anything. Same fix, no silent loss, small extra surface.
3. **Merge with the default set to the longest option (30 days).** Caps the growth bug, gives a wider margin before deletion, diverges from the documented constant.

**Recommendation: option 2.** It fixes a real 23 MB/min defect while keeping the "unknown state must mean keep, never delete" principle @thunter009 applied to the fail-open paths pointed at the user as well as the filesystem. I can implement the notice on top of this branch if you want it.

**To approve option 1 as-is:** replace `state:awaiting-human` with `state:ready-to-merge` and `risk:high` with `risk:medium`. For option 2 or 3, the label change is the same and I will push the change first.

PR #706, opened 2026-08-30
Re-resolve the calendar event at finalize against the session's real interval

thunter009:
Opening this as the follow-up invited in #701's merge comment: *"Noted your follow-up about re-resolving the event at `finalizeSession` against the finished session's real interval. Agreed that it is a separate change and that this one stands on its own — please do open it if you want to take it."*

## Root cause

The session→calendar-event binding is resolved once, at session start, and never revisited:

- `App/LiveSessionController.swift:352-356` — a manual start resolves via `container.calendarManager?.currentEvent(...)` and packs the result into `MeetingMetadata` (line 362).
- `App/MeetingDetectionController.swift:413-415` — the auto-detect path does the same (metadata built at 433-439).
- At finalize, `LiveSessionController.finalizeCurrentSession` passes `calendarEvent: endingMetadata?.calendarEvent` (line 818) — the object chosen at minute zero — into `SessionFinalizeMetadata`.

`currentEvent` picks from a ±15-minute window around the start instant (`Meeting/CalendarManager.swift:70-71`), ranked by `CalendarEventMatcher.bestMatch(among:at:)` (line 244) — proximity to a single instant. That choice can be wrong once the real interval is known: a recording that starts in the tail of the previous event binds to *that* event, because at minute zero the previous event is the closer candidate; a session that runs long may look nothing like the 15-minute block it bound to at start.

This matters more than one stale field, because three consumers read the finalize-time binding and inherit a wrong one:

1. **Speaker-name seeding** — `Storage/SessionRepository.swift:446-454` seeds speaker names from `metadata.calendarEvent`'s invitees.
2. **Per-meeting-family folder routing** — `App/LiveSessionController.swift:823-827` (`meetingFamilyPreferences(for: event)?.folderPath`).
3. **Notes template selection + `meetingFamilyKey`** — post-meeting notes load the stored binding (`LiveSessionController.swift:1117-1121`, `resolvedNotesTemplate` at 1164), and `SessionRepository.swift:1979` derives `meetingFamilyKey` from it.

## Change

- **`CalendarManager.events(overlapping:excludingCalendarIDs:)`** — fetches the non-all-day events overlapping the recording's real `[startedAt, endedAt]` interval. It only fetches; all ranking lives in `CalendarEventMatcher` as pure functions over `CalendarEvent` (the separation that made #701 testable without an `EKEventStore`).
- **`CalendarEventMatcher.bestMatch(among:overlapping:)`** — interval ranking. The key preserves #701's semantics: meeting-likeness remains the *first* key — a personal block that contains more of the recording still loses to a real meeting — then **coverage fraction** (overlap ÷ the event's own duration, quantized to five buckets), then the closest start, then the shorter event, ending in `startDate` so the ordering stays total and independent of EventKit's ordering. Coverage is a fraction rather than raw overlap seconds so a fully-covered 30-minute meeting outranks a 3-hour block the recording merely dips into, and the bucket quantization keeps one extra second of overlap from flipping the winner — near-ties fall through to start proximity.
- **`CalendarEventMatcher.finalBinding(startBinding:isUserChosen:candidates:interval:)`** — the pure finalize-time decision. The start-time binding *competes* with the challengers under the same ranking key (it is appended to the candidate pool), so a meeting-like incumbent can never lose to a non-meeting-like challenger. Zero-overlap candidates are dropped before ranking (EventKit's predicate bounds are inclusive, so a back-to-back neighbour can slip in with 0s of overlap) — but the incumbent competes regardless. A recording with *no* start-time binding only gains one when the winner is meeting-like, so an ad-hoc recording is never relocated or re-titled by the personal block that happens to contain it. A user-chosen binding is always returned unchanged.
- **Finalization** (`LiveSessionController.finalizeCurrentSession`) calls this before deriving the persisted title, and the winner feeds `SessionFinalizeMetadata`, the folder routing, and — via the stored `session.json` — notes-template resolution. The title fallback now derives from the *final* binding rather than the start-time one, so a rebound session is not persisted under the old event's name (which would have corrupted `MeetingHistoryResolver` history keys and ghost reconciliation). An explicit live topic still wins; `endingMetadata?.title` still covers sessions titled by the detected app.
- **`MeetingMetadata.calendarEventIsUserChosen`** — one design addition the original sketch didn't have. `startSession(calendarEventOverride:)` collapses a user-picked event and an auto-resolved one into the same metadata shape, so provenance has to be carried explicitly for finalize to respect an explicit choice. The flag travels with the metadata through the state machine, so it survives every start path (detail-pane "record this event" and the external command set it; manual toggle and auto-detect default to false). Decoding is explicit (`decodeIfPresent ?? false`), so any payload written before the field existed still decodes.

Re-resolution runs only when calendar integration is enabled, the user did not pick the event, and the session interval is non-empty.

## Tests

Fourteen new cases, modeled on #701's style — ranking assertions run each tie in both input orderings:

Interval ranking (`CalendarEventMatcherTests`):
- `testIntervalMatchRebindsSessionThatStartedDuringThePreviousEvent` — also asserts that the point-in-time matcher picks the *previous* event at the recording's start instant, documenting the failure mode this PR fixes.
- `testIntervalMatchKeepsFirstOfBackToBackMeetingsWhenSessionSpansBoth` — both events fully covered → buckets tie → start proximity keeps the meeting the user pressed record for.
- `testIntervalMatchPrefersContainedMeetingOverLongWorkshopOnOverrun` — meeting 1800/1800 → bucket 4 beats workshop 2100/10800 → bucket 1, where raw overlap seconds (2100 > 1800) would have picked the workshop.
- `testIntervalMatchKeepsMeetingLikenessAheadOfOverlap` — #701's first key preserved.
- `testIntervalMatchRanksPersonalBlocksByCoverageWhenNoCandidateLooksLikeAMeeting` — in-person recordings with no meeting-like candidate still land on the containing block.
- `testIntervalMatchReturnsNilWithoutCandidates`

Finalize decision (`finalBinding`, all pure):
- `testFinalBindingRespectsUserChosenEvent`
- `testFinalBindingKeepsMeetingIncumbentOverNonMeetingChallenger` — session recorded 11:50–11:58 for a 12:00 standup overlaps only a "Focus time" block; the standup is retained.
- `testFinalBindingRebindsToBetterOverlappingMeeting`
- `testFinalBindingIgnoresZeroOverlapChallengers`
- `testFinalBindingDoesNotAttachNonMeetingEventToAdHocRecording`
- `testFinalBindingAttachesMeetingEventToAdHocRecording`
- `testFinalBindingKeepsIncumbentWhenNothingOverlaps`

Plus `testMeetingMetadataDecodesPayloadWithoutUserChosenFlag` (`MeetingStateTests`) for the legacy-payload decode.

`swift build -c debug`, `swift test` (737 tests, 0 failures), and `swift build -c release` are green locally. Existing `LiveSessionControllerTests` exercise `calendarEventOverride` through full start→stop→finalize (folder routing, ghost-session merge, family template selection) and pass unchanged.

## Risk

- The finalize-time EventKit query is one predicate over the session interval — same cost class as the start-time lookup, once per session.
- Behavior changes only where the interval ranking disagrees with the minute-zero ranking *and* the incumbent loses a fair fight under the same key — a strictly-worse challenger cannot displace it. Empty candidates fall back to the start binding.
- Two adjacent behaviors are noted but deliberately unchanged:
  - **Ghost-session merge** (`Storage/SessionRepository.swift:1102-1133`) can propagate a short ghost's re-resolved binding onto the surviving session it merges into. Known, low-frequency, and bounded by the same ranking; left as is.
  - **Cloud transcription models**: `endingMetadata.startedAt` slightly predates actual audio capture, so the re-resolution interval is marginally wider than the recording. Pre-existing timestamp semantics, unchanged here.

## Revised after self-review

An adversarial pass over the first version of this branch found one real bug and several sharp edges; this revision (second commit) addresses them:

1. **The start binding now competes instead of being replaced unconditionally.** v1 used `reResolved ?? startBinding`, so *any* challenger — including a strictly worse one — displaced the incumbent: a session recorded 11:50–11:58 for a 12:00 standup overlaps only a personal "Focus time" block and would have rebound from a real meeting to a non-meeting. The incumbent now joins the candidate pool under the same ranking key, so meeting-likeness protects it (`finalBinding`, with a regression test).
2. **Raw overlap seconds → coverage-fraction buckets.** v1's absolute-overlap key had containment bias: it flipped a session spanning back-to-back meetings onto the second (longer) meeting, and let a 3-hour workshop steal a fully-covered 30-minute meeting the moment the session overran by five minutes. Ranking now uses overlap ÷ event duration quantized to buckets 0–4; the quantization also removes the one-extra-second-of-overlap flip. Two of v1's tests encoded the absolute-overlap semantics and were replaced by the scenarios above.
3. **Title derives from the final binding.** v1 still derived the persisted title from the start-time event while writing the re-resolved event, so a rebound session could persist titled "A" with `calendarEvent` B — corrupting title-based history keys. Title derivation now follows re-resolution.
4. **Zero-overlap challengers are filtered** (inclusive predicate bounds), the **fresh-binding gate** stops ad-hoc recordings from being claimed by personal blocks, and the finalize decision was **extracted into pure `CalendarEventMatcher.finalBinding`** with direct unit tests — replacing v1's "exercised indirectly" caveat.
5. **Explicit `decodeIfPresent ?? false`** for the new metadata flag, with a legacy-payload decode test.

### DIFF
--- a/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
+++ b/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
@@ -781,7 +781,22 @@ final class LiveSessionController {
         } else {
             endingMetadata = nil
         }
-        let metadataTitle = endingMetadata?.title ?? endingMetadata?.calendarEvent?.title
+        // 3b. Re-resolve the calendar binding now that the session's real
+        // interval is known. The binding made at start came from a window
+        // around minute zero and can point at the wrong event once the actual
+        // span is visible.
+        let endedAt = Date()
+        let calendarEvent = resolvedFinalizeCalendarEvent(
+            endingMetadata: endingMetadata,
+            endedAt: endedAt,
+            settings: settings
+        )
+        // Titles derive from the final binding, not the start-time one, so a
+        // rebound session is not persisted under the old event's name.
+        // `endingMetadata?.title` covers sessions with no calendar event
+        // (e.g. titled by the detected meeting app); an explicit live topic
+        // still wins below.
+        let metadataTitle = calendarEvent?.title ?? endingMetadata?.title
         let title = coordinator.transcriptStore.conversationState.currentTopic.isEmpty
             ? metadataTitle : coordinator.transcriptStore.conversationState.currentTopic
         let meetingAppName = endingMetadata?.detectionContext?.meetingApp?.name
@@ -805,17 +820,6 @@ final class LiveSessionController {
         )
         let transcriptIssue = Self.transcriptIssue(for: recordingHealthInput)
         let emptySessionClassification = Self.emptySessionDiagnosticClassification(for: recordingHealthInput)
-
-        // 3b. Re-resolve the calendar binding now that the session's real
-        // interval is known. The binding made at start came from a window
-        // around minute zero and can point at the wrong event once the actual
-        // span is visible.
-        let endedAt = Date()
-        let calendarEvent = resolvedFinalizeCalendarEvent(
-            endingMetadata: endingMetadata,
-            endedAt: endedAt,
-            settings: settings
-        )
 
         // 4. Finalize: closes file handle, backfills cleaned text, writes session.json
         await coordinator.sessionRepository.finalizeSession(
@@ -1071,9 +1075,9 @@ final class LiveSessionController {
     /// (`CalendarManager.currentEvent`), so a recording that started early can
     /// bind to the preceding event, and a meeting that ran long may look
     /// nothing like the event picked at start. Once the real interval is
-    /// known, prefer the event that best overlaps it. An event the user picked
-    /// explicitly is never second-guessed, and the start-time binding is kept
-    /// when re-resolution finds nothing.
+    /// known, that binding competes against the events overlapping the
+    /// interval under one ranking — see `CalendarEventMatcher.finalBinding`.
+    /// An event the user picked explicitly is never second-guessed.
     private func resolvedFinalizeCalendarEvent(
         endingMetadata: MeetingMetadata?,
         endedAt: Date,
@@ -1086,11 +1090,16 @@ final class LiveSessionController {
         guard let settings, settings.calendarIntegrationEnabled else { return startBinding }
         guard endedAt > endingMetadata.startedAt else { return startBinding }
 
-        let reResolved = container.calendarManager?.bestEvent(
-            overlapping: DateInterval(start: endingMetadata.startedAt, end: endedAt),
-            excludingCalendarIDs: settings.excludedCalendarIDs
+        let interval = DateInterval(start: endingMetadata.startedAt, end: endedAt)
+        return CalendarEventMatcher.finalBinding(
+            startBinding: startBinding,
+            isUserChosen: endingMetadata.calendarEventIsUserChosen,
+            candidates: container.calendarManager?.events(
+                overlapping: interval,
+                excludingCalendarIDs: settings.excludedCalendarIDs
+            ) ?? [],
+            interval: interval
         )
-        return reResolved ?? startBinding
     }
 
     private func scheduleAutoNotesIfNeeded(
--- a/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
+++ b/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
@@ -359,7 +359,10 @@ final class LiveSessionController {
             message: "Start requested (calendarEvent=\(calEvent == nil ? "no" : "yes"))"
         )
         pendingInitialScratchpad = initialScratchpad?.trimmingCharacters(in: .newlines)
-        let metadata = MeetingMetadata.manual(calendarEvent: calEvent)
+        let metadata = MeetingMetadata.manual(
+            calendarEvent: calEvent,
+            calendarEventIsUserChosen: calendarEventOverride != nil
+        )
 
         if settings.transcriptionModel.isCloud {
             state.errorMessage = nil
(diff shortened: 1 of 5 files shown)


yazins-ai:
Merged — thank you @thunter009. This is the follow-up invited in #701, and it delivers what that discussion asked for.

What makes it convincing is that you did not argue the binding is cosmetically wrong; you traced who actually reads it. Speaker-name seeding, per-meeting-family folder routing, and template selection plus `meetingFamilyKey` all inherit the start-time binding, so one bad choice at minute zero propagates into names on the transcript, the folder the export lands in, and the template the notes are written against. That is what turns a stale field into a bug worth a 572-line change.

The design decision I most want to note is keeping the fetch and the ranking apart: `events(overlapping:)` only fetches, and every ordering rule stays in `CalendarEventMatcher` as pure functions over `CalendarEvent`. That is exactly the separation that made #701 testable without an `EKEventStore`, and extending it rather than working around it is why this PR could ship with real test coverage of the ranking rather than assertions about a mocked store.

Preserving meeting-likeness as the *first* key is the right call and the one I would have pushed back on had it changed. A personal block that happens to contain more of the recording still losing to a real meeting is the property #701 established, and interval ranking could easily have quietly overturned it. Coverage as a fraction rather than raw overlap seconds is the other good detail — it is what stops a three-hour focus block from beating a fully-covered thirty-minute meeting.

Ending the sort key in `startDate` so the ordering is total and independent of EventKit's own ordering is the kind of thing that prevents a flaky test six months from now.

PR #707, opened 2026-08-30
Surface auto notes-generation failures instead of failing silently

thunter009:
## Root cause

When automatic post-meeting notes generation fails, the `catch` in `generatePostMeetingNotes` (`OpenOats/Sources/OpenOats/App/LiveSessionController.swift:1156`) only records a diagnostics breadcrumb. `DiagnosticsBreadcrumbStore.record` gates all recording on the `diagnosticLoggingEnabled` pref, which defaults to off (`OpenOats/Sources/OpenOats/Utils/DiagnosticsSupport.swift:111`). So on a stock install a failed notes call writes nothing anywhere — no retry, no notification, no state change. The meeting's notes silently never appear.

## What changed

- One retry with a 3s backoff inside `generatePostMeetingNotes`. Transient failures (local LLM server warming up, brief network drop) recover without user involvement.
- On final failure, a user notification is posted via a new `NotificationService.postNotesFailed(sessionID:)`, modeled on `postBatchCompleted` (`Meeting/NotificationService.swift:170`): same `ensurePermission()` authorization handling, static copy constants, per-session identifier, `try?` delivery. It is a no-op in unbundled runs (`isAvailable` guard), like all other notifications.
- The existing diagnostics breadcrumb is kept; a second breadcrumb records the retried first attempt.
- No new UI. The notification copy points at the existing "Generate Notes" action (`Views/ContentView.swift:615`) for manual recovery.

## Testing

- New copy test in `NotificationServiceTests`, same shape as the existing batch-completed copy test.
- `swift build -c debug`, full `swift test` (724 tests, 0 failures), and `swift build -c release` pass locally (macOS, arm64).

## Risk

Low. The success path is unchanged. The retry adds at most one extra LLM call plus 3 seconds before the existing failure handling. Notification delivery is permission-guarded and fire-and-forget.

---

## Revised after self-review

**Blocker: tapping the notification started a recording.** The delegate routed responses on the action identifier alone (`NotificationService.swift:242-244`). A tap on a notification body arrives as `UNNotificationDefaultActionIdentifier`, which fell into `default: onAccept?()` → `handleDetectionAccepted` → a new recording — while this PR's own copy asked the user to open the meeting. The same tap also cleared `pendingTimeoutTask`, so a detection prompt still on screen would never time out.

Responses now route by the *request* identifier as well: only `meeting-detection` owns the accept / not-a-meeting / ignore-app / dismiss callbacks, and a response to any other notification is a no-op that touches no state. `postBatchCompleted` had the same latent bug — a tap on "Re-transcription Complete" would also have started a recording — and the same change fixes it.

**Coverage hole: manual recordings still got silence.** `container.notificationService` was only populated by `AppContainer.enableDetection` (`:223`), which `ContentView` calls only when auto-detect is on (`:346,355`). Anyone recording manually hit the exact pre-PR behaviour this PR set out to fix. `AppContainer.alertNotificationService()` now returns the detection controller's service when there is one and creates a standalone service otherwise. `NotificationService` already no-ops on unbundled builds via its `isAvailable` guard, so nothing changes there.

**Retry only what a retry can fix.** A missing API key, a malformed base URL, a 401/404, or a model that returns nothing produce the identical failure on a second attempt — retrying only delays the notification the user needs. `isRetriableNotesFailure` now retries transport-class failures (connection refused/reset, timeout, DNS, TLS) and busy-server responses (429, 5xx); everything else notifies immediately with its specific error. Getting there meant `NotesEngine` had to stop flattening errors into a display string: it keeps the underlying error and rethrows it, so the message users see is unchanged while callers can classify.

**Backoff is 15s, not 3s.** `streamCompletion` sets a 300s idle timeout with a comment noting that local model first-token latency routinely exceeds 60s. A 3s backoff re-hits a server that has not finished warming up, which was the exact case the retry was added for.

**The retry widened two races, so the write is re-checked.** `hasNotes` was read once at the top, before a request that can run for minutes plus a 15s backoff plus a second request. The session is now re-read immediately before `saveNotes`: if the user wrote notes by hand in that window, the generated notes are discarded silently rather than clobbering them. And because `saveNotes` calls `createDirectory`, saving into a session the user deleted mid-flight would resurrect it as an empty `sessions/<id>/` shell — a new `SessionRepository.sessionExists(id:)` check prevents that, with no notification since nothing failed.

**Cancellation is not a failure.** `CancellationError` from the backoff sleep (app quitting, task torn down) is caught explicitly and aborts quietly instead of posting "Notes Generation Failed".

**Copy carries the reason.** The notification body now leads with the specific error and then the manual remedy, because "OpenRouter API key required" and "The request timed out" call for different actions. If #704 also lands, the reasoning-only / empty-response message it introduces reaches users through this body.

**Testing.** 734 tests, 0 failures (was 724). New coverage: taps and dismissals on notes-failed and batch-completed notifications route to no-ops, every detection action still routes to its callback, the reason-bearing copy and its fallback, and the retry classifier across transport codes, deterministic errors, busy-server statuses, and cancellation.

## Merge order

- **After #704.** That PR adds `OpenRouterError.reasoningOnlyResponse`. The retry classifier here is an
  allowlist of transport-class failures, so the new error becomes non-retriable automatically once #704
  lands — no follow-up edit needed, but the behaviour is only complete with both.
- **Before #708.** The two branches conflict in `NotificationService.swift` (3 hunks), `AppContainer.swift`,
  and `LiveSessionController.swift`. Both rework how notification callbacks are routed; taking this one
  first keeps the action-identifier gate as the base that #708 builds on.

---

## Follow-up review: the re-check was not atomic

**A separate actor call is still a time-of-check/time-of-use gap.** The re-check added above ran as its own `await` into `SessionRepository` — `sessionExists`, then `loadSession`, then `saveNotes`, three calls into an `actor` from a `@MainActor` method. Every one of those is a suspension point, and the main actor is free while each is in flight. `NotesController.deleteSession` (`App/NotesController.swift:914-916`) and the manual save at `:610` both reach the same repository the same way, so either can land between the check and the write. The delete case is the one the check existed to prevent: `saveNotes` calls `createDirectory`, so it recreates `sessions/<id>/` as an empty shell. The manual-save case overwrites notes the user just typed.

**Both guards now live inside the write.** `SessionRepository.saveGeneratedNotesIfAbsent(sessionID:template:markdown:generatedAt:)` checks existence, checks `hasNotes`, normalizes the heading against the session's current title and start date, and writes — one actor operation, no suspension between the check and the write. The caller switches on the returned outcome (`.written` / `.sessionMissing` / `.notesAlreadyExist`) and records the same two diagnostics breadcrumbs as before, so user-visible behaviour is unchanged in every non-racing case. Heading normalization moved into the repository because it needs the same metadata read that decides whether to write at all.

**Testing.** 738 tests, 0 failures (was 734). New `SessionRepositoryTests` coverage: the write path including the heading derived from the current title; a deleted session is not resurrected; hand-written notes survive; and a delete racing the write leaves no orphan `notes.md` in either serialization order.

### DIFF
--- a/OpenOats/Sources/OpenOats/App/AppContainer.swift
+++ b/OpenOats/Sources/OpenOats/App/AppContainer.swift
@@ -19,6 +19,21 @@ final class AppContainer {
     /// even when detection is not enabled.
     private(set) var notificationService: NotificationService?
 
+    /// Notification service for alerts that are not tied to meeting detection,
+    /// such as a post-meeting notes failure.
+    ///
+    /// `notificationService` is only populated once detection has been enabled,
+    /// so users who record manually would otherwise get no alert at all. Reuse
+    /// the detection controller's instance when there is one and create a
+    /// standalone service otherwise; `NotificationService` no-ops on its own
+    /// when notifications are unavailable (unbundled builds).
+    func alertNotificationService() -> NotificationService {
+        if let notificationService { return notificationService }
+        let service = NotificationService()
+        notificationService = service
+        return service
+    }
+
     /// Calendar manager for looking up current events.
     /// Created when the calendar integration setting is enabled.
     private(set) var calendarManager: CalendarManager?
--- a/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
+++ b/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
@@ -1121,65 +1121,156 @@ final class LiveSessionController {
         )
         let scratchpad = await coordinator.sessionRepository.loadScratchpad(sessionID: sessionID)
 
+        let generatedMarkdown: String
         do {
-            let generatedMarkdown: String
-            do {
-                generatedMarkdown = try await coordinator.notesEngine.generateMarkdownDetached(
-                    transcript: sessionData.transcript,
-                    template: template,
-                    settings: settings,
-                    calendarEvent: sessionData.calendarEvent,
-                    scratchpad: scratchpad.isEmpty ? nil : scratchpad,
-                    customGuidance: customGuidance
-                )
-            } catch {
-                // One retry after a short backoff: notes calls fail transiently
-                // (local LLM server warming up, brief network drop).
-                DiagnosticsSupport.record(
-                    category: "meeting",
-                    message: "Auto-generated post-meeting notes attempt failed for \(sessionID), retrying: \(error.localizedDescription)"
-                )
-                try await Task.sleep(for: .seconds(3))
-                generatedMarkdown = try await coordinator.notesEngine.generateMarkdownDetached(
-                    transcript: sessionData.transcript,
-                    template: template,
-                    settings: settings,
-                    calendarEvent: sessionData.calendarEvent,
-                    scratchpad: scratchpad.isEmpty ? nil : scratchpad,
-                    customGuidance: customGuidance
-                )
-            }
-
-            let notes = GeneratedNotes(
-                template: coordinator.templateStore.snapshot(of: template),
-                generatedAt: Date(),
-                markdown: GeneratedNotes.normalizedMarkdown(
-                    generatedMarkdown,
-                    title: session.title,
-                    date: session.startedAt
-                )
+            generatedMarkdown = try await generateNotesMarkdownWithRetry(
+                sessionID: sessionID,
+                transcript: sessionData.transcript,
+                template: template,
+                settings: settings,
+                calendarEvent: sessionData.calendarEvent,
+                scratchpad: scratchpad.isEmpty ? nil : scratchpad,
+                customGuidance: customGuidance
             )
-
-            await coordinator.sessionRepository.saveNotes(sessionID: sessionID, notes: notes)
-            await coordinator.loadHistory()
-
-            if coordinator.lastEndedSession?.id == sessionID {
-                coordinator.lastEndedSession = await coordinator.sessionRepository.loadSession(id: sessionID).index
-            }
-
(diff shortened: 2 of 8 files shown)


yazins-ai:

The reported bug is a good one on its own: gating the only record of a failure behind `diagnosticLoggingEnabled`, which defaults to off, means the stock configuration is precisely the one that tells the user nothing. A failure that is only visible to users who already turned on diagnostics is not a logged failure.

But the reason this PR is more valuable than its title is the blocker you found in your own change. Routing notification responses on the action identifier alone meant a tap on the notification *body* arrived as `UNNotificationDefaultActionIdentifier` and fell through `default:` into `onAccept` — starting a recording. Your new notification asked the user to open the meeting, so following that instruction would have started recording them. On a privacy-first recorder that is the worst possible failure mode, and it was latent on `postBatchCompleted` before this PR: a tap on "Re-transcription Complete" would have done the same thing. The same tap also cleared `pendingTimeoutTask`, so a detection prompt still on screen would never time out.

Routing by request identifier as well, so only `meeting-detection` owns the accept and dismiss callbacks and every other notification is inert, is the right shape — it makes the class of bug impossible rather than fixing this instance.

The second catch is nearly as good: `container.notificationService` was only populated by `AppContainer.enableDetection`, which `ContentView` calls only when auto-detect is on. So anyone recording manually got exactly the silence this PR set out to remove. Shipping the fix without that would have left the feature not working for a large group of users while appearing to work.

The retry is well-judged — one attempt, three seconds, and an honest account of the cost as one extra LLM call before the existing failure handling.

PR #708, opened 2026-08-30
Add opt-in auto-record for detected meetings

thunter009:
## Problem

Meeting auto-detect never auto-records. The chain today:

- `MeetingDetector` emits `.detected` → `MeetingDetectionController.handleMeetingDetected` posts a notification (`Sources/OpenOats/App/MeetingDetectionController.swift:384`)
- `NotificationService.postMeetingDetected` arms a 60-second timeout that removes it (`Sources/OpenOats/Meeting/NotificationService.swift:154-164`)
- Recording starts only if the user taps within that window (`handleDetectionAccepted`, `MeetingDetectionController.swift:402`)

Step away from the desk, miss the banner → the meeting is lost. For people replacing tools that record reliably (e.g. Granola), this is the biggest capture-reliability gap.

## Change

New setting **"Automatically record detected meetings"**, default **off** — stock behavior is untouched.

When enabled, `handleMeetingDetected` invokes the accept path directly instead of posting the actionable prompt, then posts an informational "Recording Started" notification so the user knows a recording is in progress.

- All existing pre-checks still run before the auto-accept: already-recording (`isSessionActive`), per-session dismissals (`dismissedEvents`), permanently ignored apps (`ignoredAppBundleIDs`). The yielded `.accepted` event flows through the same coordinator path as a user tap, so downstream behavior (silence monitoring, app-exit monitoring, state machine) is identical.
- The informational notification has no actions and no 60-second timeout/auto-removal. It uses a distinct identifier, and `userNotificationCenter(_:didReceive:)` ignores taps on it, so the default-tap-means-accept behavior cannot re-enter the accept path.

## Risk and why default-off

Honest trade-off: a false-positive detection now starts a real recording, with audio retention if "Save audio recording" is on. The tap-to-start prompt is a consent gate as much as a UX step, and this setting removes it. That is the right trade only for users who deliberately opt in to Granola-style capture reliability, so it ships opt-in and default-off, with the toggle directly under the detection toggle and copy that says exactly what it does. The existing "No audio or video is captured until you accept the notification" copy stays true for the default configuration; happy to make it conditional if you prefer.

## Also in this PR

`NotificationService.isAvailable` now also detects the XCTest runner (`NSClassFromString("XCTestCase") == nil`). Under `swift test` the xctest tool has a bundle identifier but no bundle proxy, so `UNUserNotificationCenter.current()` raises `NSInternalInconsistencyException` the moment anything constructs `NotificationService` — which is why no existing test exercises `MeetingDetectionController.setup`. With the guard, `setup` is testable; the new controller tests rely on it.

## Tests

- `MeetingDetectionControllerTests`: auto-record yields `.accepted` with no notification tap (camera-triggered); default-off yields nothing without a tap; already-recording suppresses auto-accept
- `SettingsStoreTests`: default false + round trip

`swift build -c debug` and `-c release` clean; full suite 728 tests, 0 failures (5 new).

---

## Revised after self-review

Re-reading this with fresh eyes turned up two things I would not want to ship, plus a handful of smaller ones.

**Auto-record could start capture with no notification at all (blocker).** The old order was: accept first, then post the informational notification fire-and-forget. But the post begins with a permission guard, so a denied or revoked notification permission — or notifications being unavailable — produced a real recording with nothing on screen to say so. That is the exact inversion of the manual path, which fails safe to *no* recording. `postAutoRecordingStarted` now reports whether delivery succeeded, and the detection controller posts *before* it accepts. If the notification cannot be delivered, detection falls back to the ordinary tap-to-confirm prompt. Both branches are covered by tests.

**The consent copy had become false (blocker).** "No audio or video is captured until you accept the notification" is simply untrue once auto-record is on. The detection caption is now conditional on the toggle, the explanation sheet states the auto-record case, and the toggle's own caption says plainly: "Starts recording without asking. A notification is posted when recording begins."

Also in this revision:

- **Consent gate.** Every human-initiated start checks `hasAcknowledgedRecordingConsent`; the auto path checked nothing. It now requires the same acknowledgement, and falls back to the prompt when it has not been given.
- **Monitor arming.** The silence and app-exit monitors were armed unconditionally *before* `handle(...)`. Since `.ending` + `.userStarted` is a no-op transition, that left orphan monitors alive into `.idle`, where a stale silence timer would stop a later manual session. Monitors are now armed only after the state machine actually transitions. **This pre-arming is a pre-existing defect on `main`, not something this PR introduces — but auto-record makes it far easier to hit, so it is fixed here rather than widened.**
- **Session check widened.** The detection controller treated only `.recording` as an active session. It now treats any non-idle state as active, so nothing auto-starts while a previous session is still finalizing.
- **Manual-start race.** During a cloud-model preflight the coordinator is still `.idle` while the start is already committed, so auto-record could steal it and attach detection metadata plus auto-stop monitors to the user's own session. There was a clean seam for this — the preflight task is nil-checked in a `defer` — so it is exposed as `isStartInFlight` and folded into the same session check rather than being papered over.
- **Stale detection banner.** The auto path never removed a previously delivered detection banner, so a tap on it could re-enter the accept path mid-session. It is now cancelled and removed before accepting.
- **The informational notification is removed when the meeting ends** instead of sitting in Notification Centre indefinitely.
- **No more `XCTestCase` string lookup in production.** `NotificationService` takes an injectable availability flag instead, and the tests pass their own. Worth noting why the guard existed at all: no test on `main` ever called `MeetingDetectionController.setup`, so nothing had previously constructed the service under the xctest runner, where `UNUserNotificationCenter` raises `NSInternalInconsistencyException` (bundle identifier present, bundle proxy nil). The new controller tests do exercise `setup`, hence the seam.

**Interaction with #709.** That PR lets you add arbitrary bundle IDs to the detection list; this one removes the confirmation step. Together they mean a custom browser entry can make any microphone-using website start a real recording, so the auto-record caption now says exactly that.

Left deliberately unaddressed, flagged for your call:

- **Launch at login plus auto-record.** `evaluateImmediate` bypasses the 5-second debounce, so a login while a call is already running can begin capture more or less at login. That is arguably the intended behaviour of both settings combined, but it is the least expected consequence of ticking two boxes, and I would rather have your read on it than pick a default myself.
- **Microphone permission denied** leaves the state machine claiming `.recording` while nothing is captured. Pre-existing on `main`; auto-record makes it less visible, because there is no longer a user action to correlate with the silence. Out of scope here.
- **The toggle is retained across detect cycles** by design — it is a setting, not a per-meeting choice.

Full suite: 730 tests, 0 failures (up from 728, two new).

## Merge order

- **After #707.** The two branches conflict in `NotificationService.swift` (3 hunks), `AppContainer.swift`,
  and `LiveSessionController.swift`. #707 introduces the action-identifier gate that this PR's
  notification changes sit on top of.
- **Before #709.** The two branches conflict in `SettingsStore.swift` (3 hunks), `SettingsView.swift`, and
  `MeetingDetectionController.swift`.

---

## Follow-up review: teardown left the notification behind

**"Removed when the meeting ends" had one hole: detection being switched off mid-meeting.** `clearAutoRecordingStarted()` was reachable only from `handleMeetingEnded()` (`App/MeetingDetectionController.swift:425`), which rides the `.ended` event off the detector. `teardown()` cancels the detection event loop and the app-exit monitor before it stops the detector (`:201-213`), and `AppContainer.disableDetection` stops the coordinator's event loop first (`App/AppContainer.swift:236`), so `.ended` never arrives and `handleMeetingEnded()` never runs. Teardown then drops the service reference (`:217`), leaving nothing that can retract the banner. Since the informational notification deliberately carries no timeout, "Recording Started" stayed in Notification Center indefinitely.

`teardown()` now clears it next to `cancelPending()`, before the reference is dropped.

**Testing.** 731 tests, 0 failures (was 730). The new `MeetingDetectionControllerTests` case auto-records a camera-triggered detection, tears the controller down while it is still active, and asserts the clear happened; it fails on the previous commit with the same assertion message.

### DIFF
--- a/OpenOats/Sources/OpenOats/App/AppContainer.swift
+++ b/OpenOats/Sources/OpenOats/App/AppContainer.swift
@@ -230,7 +230,12 @@ final class AppContainer {
         let controller = MeetingDetectionController()
         controller.isSessionActive = { [weak coordinator] in
             guard let coordinator else { return false }
-            return coordinator.isRecording
+            // Any non-idle state counts: prompting (or auto-starting) while a
+            // previous session is still finalizing (.ending) must be suppressed,
+            // as must a start while a cloud-model preflight is in flight (the
+            // coordinator is still .idle then, but the start is committed).
+            return coordinator.state != .idle
+                || coordinator.liveSessionController?.isStartInFlight == true
         }
         detectionController = controller
         controller.setup(settings: settings)
--- a/OpenOats/Sources/OpenOats/App/AppCoordinator.swift
+++ b/OpenOats/Sources/OpenOats/App/AppCoordinator.swift
@@ -317,6 +317,13 @@ final class AppCoordinator {
                 guard let self, !Task.isCancelled else { break }
                 switch event {
                 case .accepted(let metadata):
+                    // Let the state machine decide first. Arming the auto-stop
+                    // monitors before a vetoed transition (.ending + .userStarted
+                    // is a no-op) leaves orphan monitors that survive into .idle
+                    // and kill a later manual session with a stale silence timer.
+                    let stateBefore = self.state
+                    self.handle(.userStarted(metadata), settings: self.activeSettings)
+                    guard self.state != stateBefore else { break }
                     let signal = metadata.detectionContext?.signal
                     if case .appLaunched(let app) = signal {
                         controller.startSilenceMonitoring()
@@ -327,7 +334,6 @@ final class AppCoordinator {
                             controller.startAppExitMonitoring(bundleID: app.bundleID)
                         }
                     }
-                    self.handle(.userStarted(metadata), settings: self.activeSettings)
                 case .meetingAppExited:
                     if case .recording(let meta) = self.state {
                         let signal = meta.detectionContext?.signal
--- a/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
+++ b/OpenOats/Sources/OpenOats/App/LiveSessionController.swift
@@ -182,6 +182,9 @@ final class LiveSessionController {
 
     private var downloadTask: Task<Void, Never>?
     private var startPreflightTask: Task<Void, Never>?
+    /// True while a cloud-model preflight is validating a manual start.
+    /// The coordinator is still `.idle` then, but the start is committed.
+    var isStartInFlight: Bool { startPreflightTask != nil }
     private var scratchpadSaveTask: Task<Void, Never>?
     private var pendingInitialScratchpad: String?
 
--- a/OpenOats/Sources/OpenOats/App/MeetingDetectionController.swift
+++ b/OpenOats/Sources/OpenOats/App/MeetingDetectionController.swift
@@ -394,6 +394,20 @@ final class MeetingDetectionController {
             Log.meetingDetection.info("Detected: \(app?.name ?? "unknown", privacy: .public) (trigger: \(isCameraTrigger ? "camera" : "mic+app", privacy: .public))")
         }
 
+        // Auto-record: skip the notification round-trip and invoke the accept
+        // path directly. All pre-checks above (already recording, dismissed,
+        // permanently ignored) have passed at this point.
+        if activeSettings?.autoRecordDetectedMeetings == true {
+            if activeSettings?.detectionLogEnabled == true {
+                Log.meetingDetection.info("Auto-record enabled, accepting detection without prompt")
+            }
+            handleDetectionAccepted()
+            await notificationService?.postAutoRecordingStarted(
+                appName: isCameraTrigger ? nil : app?.name
+            )
+            return
+        }
+
         // Don't attribute camera-triggered detections to a background meeting app —
         // the camera could have been activated by a different app (e.g. browser for Google Meet).
         let posted = await notificationService?.postMeetingDetected(
--- a/OpenOats/Sources/OpenOats/App/MeetingDetectionController.swift
+++ b/OpenOats/Sources/OpenOats/App/MeetingDetectionController.swift
@@ -221,6 +221,12 @@ final class MeetingDetectionController {
         }
 
         notificationService?.cancelPending()
+        // Detection can be switched off mid-meeting. The detection event loop
+        // is already cancelled above, so `.ended` will never arrive and
+        // handleMeetingEnded() will never run — without this, a delivered
+        // "Recording Started" notification would sit in Notification Center
+        // for good.
+        notificationService?.clearAutoRecordingStarted()
         notificationService = nil
 
         if let observer = sleepObserver {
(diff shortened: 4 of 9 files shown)


yazins-ai:
Thanks @thunter009. The capture-reliability gap you describe is real — the 60-second notification window does lose meetings when someone steps away, and the implementation here is careful: default-off, all existing pre-checks retained, the informational notification given a distinct identifier so a tap cannot re-enter the accept path.

Labeled `risk:high` and `state:awaiting-human`, and this one is not about code quality.

Repo policy treats changes to the recording consent flow as maintainer-only by default, and you identified precisely why in your own risk section: the tap-to-start prompt is a consent gate, not just a UX step. This setting removes it. That is a deliberate product decision about what OpenOats promises — a privacy-first local recorder where nothing is captured until the user accepts — and it is not mine to make autonomously, however well the opt-in is built.

Two specifics for @yazinsai when reviewing:

- The existing copy "No audio or video is captured until you accept the notification" becomes false for users who enable this. You offered to make it conditional; I think that is required rather than optional if this lands, since the sentence is the product promise in miniature.
- A false-positive detection now starts a real recording, with audio retained if "Save audio recording" is on. Worth deciding alongside #705, which is about how long that retained audio lives — the two interact.

The `NotificationService.isAvailable` XCTest fix is a genuinely useful side-catch and would stand on its own in a separate PR if the auto-record decision takes a while.

thunter009:
Rebased on current `main`; it was conflicting and is not now. `swift test` — 807 tests, 0 failures.

**The consent copy is conditional, as you required.** It was already in the branch when you reviewed, so this is a pointer rather than a change. The detection caption switches on the setting:

```swift
Text(settings.autoRecordDetectedMeetings
    ? "...Automatic recording is on, so recording starts as soon as a meeting is detected."
    : "...No audio or video is captured until you accept the notification.")
```

The explanation sheet carries the same split: "No audio is captured or recorded until you accept — or, if automatic recording is enabled, until a meeting is detected (a notification is posted)." The promise sentence is never shown to a user for whom it is false.

**The `NotificationService.isAvailable` catch has landed upstream independently.** `main` now has the bundle-identifier-and-`.app` check as a stored property, so there is nothing left to split out. What remains on this branch is only making it injectable through `init` so tests can construct the service without a real bundle, which is not separable from the tests here.

The rebase also let me drop one thing: I had a redundant guard against the auto-record notification identifier in the delegate. Your `Self.route(...)` already returns `.none` for any request that is not `detectionRequestID`, so the guard was checking a case that could not reach it. One check, in one place.

On the product decision itself — that is yours and @yazinsai's, and I have nothing to add to your framing. The tap-to-start prompt is a consent gate and this setting removes it. Your point that it interacts with #705 is right: a false positive under auto-record produces a real recording, and #705 decides how long that audio then lives.

yazins-ai:
Merged with maintainer approval. Thanks @thunter009. Routing auto-accept through the same `.accepted` event as a user tap was the right call. It means silence monitoring, app-exit handling and the dismiss/ignore checks all apply without a second code path to keep in sync. Default-off kept this safe to approve.

PR #709, opened 2026-08-30
Surface custom meeting apps in Settings and apply edits live

thunter009:
## Problem

Microphone-based meeting detection only fires for a known meeting app: `handleMicSignal` guards on `scanForMeetingApp() != nil` (`Sources/OpenOats/Meeting/MeetingDetector.swift:333-334`), and the known list is the 12 hardcoded bundle IDs in `defaultMeetingApps` (`MeetingDetector.swift:379-392`). The only browser-shaped entry is the Google Meet PWA (`:390`). So a Meet/Teams/Zoom-web call in a normal browser tab is invisible to the mic path; the camera path (`MeetingDetector.swift:277-281`) is the only rescue, and camera-off browser calls are undetectable.

The plumbing to extend the list actually already exists — `SettingsStore.customMeetingAppBundleIDs` is wired into the detector via `MeetingDetectionController.setup` (`Sources/OpenOats/App/MeetingDetectionController.swift:131-133`) — but no UI reads or writes it. Today the only way to use it is `defaults write`, and even then the change needs an app relaunch because the detector snapshots the list at setup.

## Change

- **Settings UI**: a "Custom Meeting Apps" section next to the detection settings (monospaced list with remove buttons, plus an add field), shown while auto-detect is enabled. Mirrors the existing "Ignored Apps" section.
- **Normalization**: `customMeetingAppBundleIDs` is now trimmed and deduplicated on set and on load, reusing `normalizedIdentifierList` — the same treatment `excludedCalendarIDs` already gets.
- **Live application**: `ContentView` restarts detection when the list changes, guarded to run only while auto-detect is on and no recording is active (tearing down mid-recording would cancel the silence and app-exit monitors of the live session).
- `MeetingDetector.knownBundleIDs` loses `private` so tests can assert the merged set.

## Why browsers are not added to the default list

Deliberately not done, and worth being explicit about: mic-based detection would then fire for any website using the microphone — dictation, voice notes, browser games — a false-positive storm for every browser user. User extensibility is the right shape: people who live in browser meetings opt in deliberately, with the trade-off explained in the section's caption.

## Trade-offs

- Editing the list restarts detection; a pending detection notification at that moment is cancelled. Acceptable: the user is actively editing detection settings.
- Bundle IDs are free text. A typo simply never matches; entries are shown verbatim so mistakes are visible and removable.
- Detection restart is skipped during a recording, so an edit made mid-recording applies when detection next restarts (recording end toggles or next launch).

## Tests

- `MeetingDetectorTests`: merged set contains defaults + user additions; duplicates of defaults, repeated custom entries, and the app's own bundle ID all collapse/are excluded
- `SettingsStoreTests`: normalization on set and on load

`swift build -c debug` and `-c release` clean; full suite 727 tests, 0 failures (4 new).

---

## Revised after self-review

Going back over this, the headline problem was that the motivating use case did not actually work.

**Adding an app mid-call did nothing.** The restart on list change rebuilt the detector but never re-evaluated the current microphone and camera state, and a fresh `CoreAudioSignalSource` only emits on transitions. So adding your browser while the Meet call is already running changed nothing until the microphone next cycled — which, in a call, is when the call ends. The restart now calls `evaluateImmediate()`, the same way the auto-detect toggle already did.

**Case variants were dead entries.** Bundle identifiers are matched case-insensitively in practice, but the list deduplicated on the exact string and detection compared with an exact `Set` lookup. So `us.Zoom.xos` persisted, rendered as its own row, and never matched anything — a setting that looks applied and silently is not. The stored list now deduplicates without regard to case, keeping whichever spelling you typed first, and both the detection set and every lookup are case-folded. Calendar identifiers are opaque and keep their existing case-sensitive normalization; there is a test pinning that distinction.

**Edits made while recording were dropped forever.** The old guard skipped the restart mid-recording and nothing ever re-applied it, so the change quietly survived only as far as the next relaunch. The restart is now deferred and applied on the next return to idle.

**Restarting wiped this session's "Not a Meeting" dismissals.** Teardown clears them, which is right when you turn detection off, but wrong for a list edit — that is not a new session, and re-prompting for something the user already dismissed is the annoying failure. Dismissals now carry across a restart-for-list-edit and are still cleared on an explicit disable.

**Adding a built-in app looked like it worked and could not be undone.** You could add `us.zoom.xos` as a custom entry, see it in the list with a remove button, remove it, and detection would carry on regardless. Those additions are now refused with inline "already built in" feedback.

**The merge is now a pure function.** `resolveKnownBundleIDs(defaults:custom:selfID:)` handles bundled defaults plus user additions minus the app's own identifier, so `knownBundleIDs` goes back to being private and the tests assert the merge directly rather than reaching into the actor.

**Interaction with #708.** That PR adds an opt-in "automatically record detected meetings" setting. With both landed, a custom browser entry here means any microphone-using website can start a real recording with no confirmation step, so #708's caption calls that out explicitly. This PR's own caption already warns why browsers are not built in.

Possible follow-up, not done here: an `NSOpenPanel` app picker so the bundle identifier can be chosen from an installed app rather than typed. Free-text entry is honest about what it is — a typo simply never matches, and entries render verbatim — so this seemed like polish for a later pass rather than something to bundle in.

Full suite: 731 tests, 0 failures (up from 727, four new).

## Follow-up: teardown never stopped the detector

A review pass flagged a race here: `disableDetection` schedules `teardown`'s
asynchronous `meetingDetector?.stop()` and clears the property, so the task
could stop the *replacement* detector. That one cannot happen.
`disableDetection` also drops the controller itself (`AppContainer.swift:231-236`)
and `enableDetection` builds a new one, so the task's `self` is the old
controller and the replacement lives on a different object.

The premise underneath it is real, though, and worse: the property is already
nil when the task body runs. `teardown` read `meetingDetector` inside the task,
which runs after the method returns and therefore after `meetingDetector = nil`
on the next line, so the stop always applied to nil. No detector was ever
stopped. Its mic and camera monitors stay alive for the life of the process,
and this branch makes detection restarts routine — one abandoned detector per
custom meeting-app edit.

`teardown` now captures the detector before clearing the property.

`MeetingDetectionControllerTests.testTeardownStopsTheDetectorItStarted` drives a
detector into an active detection through the controller, tears down, and
asserts the detector stopped. It fails on the previous commit.

Reaching that path from a test needed one more fix. `NotificationService`'s
availability probe only ruled out a missing bundle identifier, which the
`xctest` runner does have, so constructing one raised inside
`UNUserNotificationCenter` ("bundleProxyForCurrentProcess is nil"). The probe
now requires an `.app` bundle, which is what its comment always described.

Full suite: 732 tests, 0 failures (up from 731, one new).

## Merge order

Merge after #708. The two branches conflict in `SettingsStore.swift` (3 hunks), `SettingsView.swift`, and
`MeetingDetectionController.swift` — both add settings-backed lists and restart-on-change handling to the
detection controller.

This PR also touches `ContentView.swift`, which #712 edits as well, but those two do not collide.

### DIFF
--- a/OpenOats/Sources/OpenOats/App/MeetingDetectionController.swift
+++ b/OpenOats/Sources/OpenOats/App/MeetingDetectionController.swift
@@ -208,10 +208,17 @@ final class MeetingDetectionController {
         appExitMonitorTask?.cancel()
         appExitMonitorTask = nil
 
-        Task {
-            await meetingDetector?.stop()
-        }
+        // Capture the detector first: the task body reads its captures when it
+        // runs, which is after this method returns, so reading the property
+        // there would find the nil below and stop nothing. The abandoned
+        // detector would keep its mic and camera monitors alive for the life of
+        // the process — once per detection restart, and editing the custom
+        // meeting-app list restarts detection.
+        let detector = meetingDetector
         meetingDetector = nil
+        Task {
+            await detector?.stop()
+        }
 
         notificationService?.cancelPending()
         notificationService = nil
--- a/OpenOats/Sources/OpenOats/App/MeetingDetectionController.swift
+++ b/OpenOats/Sources/OpenOats/App/MeetingDetectionController.swift
@@ -231,6 +231,14 @@ final class MeetingDetectionController {
         Log.meetingDetection.info("Detection system stopped")
     }
 
+    /// Re-seed dismissals carried across a detector rebuild. Editing the custom
+    /// meeting-app list rebuilds the detector, but that is not a new session:
+    /// the "Not a Meeting" choices the user already made must survive it.
+    /// Turning auto-detect off is a real stop, and `teardown` still clears them.
+    func restoreDismissedEvents(_ events: Set<DismissKey>) {
+        dismissedEvents.formUnion(events)
+    }
+
     // MARK: - Sleep Observer
 
     private func installSleepObserver() {
--- a/OpenOats/Sources/OpenOats/Meeting/MeetingDetector.swift
+++ b/OpenOats/Sources/OpenOats/Meeting/MeetingDetector.swift
@@ -157,7 +157,9 @@ actor MeetingDetector {
     private let knownApps: [MeetingAppEntry]
     private let customBundleIDs: [String]
     private let selfBundleID: String
-    private let knownBundleIDs: Set<String>
+    /// Resolved detection set: bundled defaults plus user additions, minus the
+    /// app's own bundle ID. Non-private so tests can assert the merge.
+    let knownBundleIDs: Set<String>
 
     /// Set to true once detection is confirmed.
     private(set) var isActive = false
--- a/OpenOats/Sources/OpenOats/Meeting/MeetingDetector.swift
+++ b/OpenOats/Sources/OpenOats/Meeting/MeetingDetector.swift
@@ -157,9 +157,8 @@ actor MeetingDetector {
     private let knownApps: [MeetingAppEntry]
     private let customBundleIDs: [String]
     private let selfBundleID: String
-    /// Resolved detection set: bundled defaults plus user additions, minus the
-    /// app's own bundle ID. Non-private so tests can assert the merge.
-    let knownBundleIDs: Set<String>
+    /// Resolved detection set, case-folded. Built by `resolveKnownBundleIDs`.
+    private let knownBundleIDs: Set<String>
 
     /// Set to true once detection is confirmed.
     private(set) var isActive = false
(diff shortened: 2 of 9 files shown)


yazins-ai:

This is the best kind of feature PR: it ships almost no new capability. `customMeetingAppBundleIDs` was already wired from `SettingsStore` through `MeetingDetectionController.setup` into the detector — the setting existed, was respected, and was reachable only by `defaults write` followed by a relaunch. Finding dead-ended plumbing and connecting it is a much smaller change than the problem it solves, and it fits the product guardrail of preferring the smallest change that works.

The section I want to single out is "Why browsers are not added to the default list". Adding a couple of browser bundle IDs to `defaultMeetingApps` would have looked like the obvious fix and would have been the wrong one — mic-based detection would then fire on dictation, voice notes, and browser games, producing a false-positive storm for every browser user. Choosing user extensibility over a default that is right for some users and actively bad for the rest is the correct trade, and putting the reasoning in the PR rather than leaving it implicit is what let me merge it without relitigating the decision.

Restarting detection on edit rather than requiring a relaunch is what makes the setting genuinely usable, and guarding the restart so it cannot happen mid-recording is the detail that makes it safe — tearing down the detector during a live session would have cancelled the silence and app-exit monitors. Deferring an edit made mid-recording is the right trade-off and you documented it.

Treating free-text bundle IDs as self-correcting — a typo simply never matches, entries render verbatim, removal is one click — is a sound reason not to build validation here.

PR #710, opened 2026-08-30
Retry notes-folder mirrors that fail while the folder is unreachable

thunter009:
## Root cause

`SessionRepository.performMirror` (`OpenOats/Sources/OpenOats/Storage/SessionRepository.swift:1942` on main) writes the notes-folder mirror via `MarkdownMeetingWriter.write`, which catches its own write error and returns nil (`Intelligence/MarkdownMeetingWriter.swift:95`). The caller swallows that with `guard let outputTarget else { return }`, and `scheduleMirror` is a fire-and-forget `Task.detached` from 7 event-triggered call sites. Nothing records the failure and nothing re-drives it. With the notes folder on a network mount that is temporarily unreachable (e.g. SMB share down), the meeting's export is silently never written.

## What changed

- `performMirror` now returns `Bool`: `false` means the export failed and should be retried; an empty transcript ("nothing to mirror") still counts as success.
- The detached mirror task reports its outcome back to the actor (`mirrorDidFinish`), which tracks pending sessions in memory and persists an optional `mirrorPending` flag in `session.json` (`SessionMetadata`) so a failed export survives an app relaunch. Metadata is only written when the flag actually changes, so steady-state success paths never touch `session.json`.
- Re-drive points for pending mirrors:
  - repository init: scans session metadata for persisted flags (alongside the existing `cleanupExpiredRetainedBatchAudio`), then retries;
  - `setNotesFolderPath`: the folder is configured after init, so this is the first moment an init-discovered pending mirror can actually run;
  - piggybacked on every `scheduleMirror`, so a pending export heals on the next meeting event once the mount is back.
- Success behavior is unchanged: same write path, same output.

## Testing

Two new `SessionRepositoryTests` (same polling shape as the existing mirror tests):

1. A regular file occupies the notes-folder path, so every mirror write fails → `mirrorPending: true` is persisted in `session.json`; pointing the notes folder at a writable directory re-drives the export and clears the flag.
2. The blocked path recovers in place (file removed, directory created — no reconfiguration): the next `saveNotes` mirror piggybacks the stuck session; both exports appear.

`swift build -c debug`, full `swift test` (725 tests, 0 failures), and `swift build -c release` pass locally (macOS, arm64). Not verified against a real SMB mount; the failure is simulated by making the target path unwritable, which exercises the same nil-return path in `MarkdownMeetingWriter.write`.

## Risk

Low. Retries only run on existing events (init, folder configuration, scheduled mirrors) — no timers or loops. A deleted session with a pending flag self-clears on retry (empty transcript reads as nothing-to-mirror success). The new metadata field is optional, so existing `session.json` files decode unchanged.

---

### Retry work is now bounded

The first cut re-drove every pending mirror from `setNotesFolderPath` — which `startTranscription` calls at every recording start (`LiveSessionController.swift:676,684`) — and from every `scheduleMirror`, with no guard against an attempt still running. Against a permanently unreachable mount the pending set only grew, and each pass spawned one more detached task doing blocking file I/O, each able to sit on the mount's own timeout.

A new `MirrorAttemptLedger` (`Storage/MirrorAttemptLedger.swift`) enforces:

- **one in-flight attempt per session** — a mirror requested while one is running is coalesced into a single re-run once that attempt reports (the markdown is already on disk, so nothing is lost);
- **a 5-minute failure backoff per session** — a session whose last attempt failed is not retried before then;
- the retry pass drives only sessions that pass both checks.

Re-applying the *same* notes folder no longer clears that backoff; only a genuine change of destination does, since past failures say nothing about a new folder.

**Resulting bound:** with a permanently unreachable notes folder, retry work is at most one blocked attempt per pending session per 5 minutes, however many recordings are started. Before, it was one attempt per pending session per recording start, unbounded and overlapping.

### Ordering race

Attempts now carry a monotonically increasing id, and `mirrorDidFinish` ignores reports from superseded attempts. Without it a slow success could clear the pending flag set by a newer attempt that actually failed, dropping the durable retry silently. This is reachable: deleting a session drops its in-flight reservation (below), so a fresh attempt can start while the old one is still blocked.

### Two writers were dropping the flag

`finalizeSession` and the post-batch metadata refresh hand-construct `SessionMetadata`, so both cleared a persisted `mirrorPending: true` — losing the retry permanently if the app quit before the next report. Both now carry the field across. The post-batch refresh drops `speakerNames` and `customNotesGuidance` in exactly the same way; that is the same defect class, flagged here rather than fixed, to keep this PR to one concern.

### Smaller fixes

- The startup scan is now `nonisolated` and runs off the actor, hopping back with the resulting ID set. It decoded one JSON file per session directory while actor-isolated, queueing every other repository call behind it at launch.
- The scan reads only canonical `session_` directories, matching the batch-audio cleanup scan, so `.recently-deleted` and stray files are skipped.
- Deleting a session, or moving it to Recently Deleted, drops its retry state instead of leaving it in the pending set.
- A failed asset copy in `synchronizeMirroredAssets` now fails the mirror, so it is retried. A *missing* referenced asset is still skipped and does not fail the mirror, because retrying would never bring it back.

### Behaviour worth knowing

**Folder split on retry.** If the notes folder is changed while mirrors are pending, those pending sessions mirror into the **new** folder while their already-exported siblings stay in the old one. This is intended: the new folder is the user's current choice, and writing to a folder the user has moved away from would be worse.

**Interaction with #707.** If both land, a retried notes generation calls `saveNotes` → `scheduleMirror`, which also runs a retry pass — so the pass runs about twice as often per meeting. It is idempotent, and now bounded by the in-flight guard and the backoff, so the extra frequency costs nothing.

### Tests

`MirrorAttemptLedgerTests` covers the in-flight guard, the backoff window, backoff clearing on a folder change, delete pruning, request coalescing, and the ordering race: attempt 1 starts, attempt 2 completes first and fails, and attempt 1's late success is rejected with the pending flag intact. The ledger is a plain value type, so that race is deterministic rather than timing-dependent.

Both existing polling tests are kept; the test repository is constructed with a short backoff so the retry path still runs. Full `swift test` is 735 tests, 0 failures (was 725); `swift build` clean (macOS, arm64).

## Follow-up: a destination change under a running attempt

A review pass found a gap in the ledger. `retryPendingMirrors` skips a session
whose attempt is still running (`MirrorAttemptLedger.swift:79`), so when the
notes folder changed while an attempt was in flight, that session was never
driven to the new folder:

- if the running attempt succeeded against the **old** folder, `mirrorDidFinish`
  cleared the pending state and the durable flag, and the new folder never
  received the export at all;
- if it failed, that failure was recorded as the session's backoff, holding the
  new folder off for a full retry interval although nothing had been tried
  there.

`setNotesFolderPath` now tells the ledger that the destination changed, and
every attempt running at that moment is marked stale. A stale attempt's report
keeps the session pending, records no backoff, and re-drives the mirror against
the new folder. `finishAttempt` returns `recorded` / `ignored` /
`staleDestination` instead of a Bool, so the caller can tell "not mine" from
"not this folder".

**The bound is unchanged.** Nothing starts an attempt while one is running, so a
session still has at most one in flight, and the re-run waits for the running
attempt to report. A permanently unreachable folder still costs at most one
blocked attempt per pending session per backoff interval; a destination change
adds at most one further attempt per session, sequentially, never overlapping.

Four new `MirrorAttemptLedgerTests`: a stale success keeps the session pending,
a stale failure leaves it immediately retryable, repeated destination changes
cannot stack attempts onto a blocked mount, and an attempt started after the
change is recorded normally. Full suite: 739 tests, 0 failures (was 735).

## Merge order

Merge after #705. Both branches edit `SessionRepository.swift` and conflict there (3 hunks plus tests):
#705 adds the audio-retention sweep, this PR adds the mirror-retry ledger.

### DIFF
--- a/OpenOats/Sources/OpenOats/Storage/MirrorAttemptLedger.swift
+++ b/OpenOats/Sources/OpenOats/Storage/MirrorAttemptLedger.swift
@@ -1,0 +1,121 @@
+import Foundation
+
+/// Bookkeeping that bounds notes-folder mirror work.
+///
+/// A mirror writes into a user-chosen folder that may live on a network mount.
+/// When that mount is unreachable, a single attempt can block for the mount's
+/// own timeout — tens of seconds for SMB. Retries are driven by ordinary app
+/// activity (every recording start re-applies the notes folder), so without
+/// bookkeeping the outstanding work for a permanently dead mount grows without
+/// bound.
+///
+/// The ledger imposes two limits, which together bound the cost of a dead mount
+/// to one blocked attempt per pending session per `retryBackoff`:
+///
+/// - **In-flight guard.** A session with a running attempt never starts a
+///   second one. A request that arrives during an attempt is coalesced into a
+///   single re-run.
+/// - **Failure backoff.** A session whose last attempt failed is not retried
+///   again until `retryBackoff` has elapsed.
+///
+/// Attempts carry a monotonically increasing id so a report can be matched to
+/// the attempt that produced it. A report from a superseded attempt is ignored
+/// — otherwise a slow success could clear the pending flag of a newer attempt
+/// that actually failed, silently dropping the durable retry.
+struct MirrorAttemptLedger: Sendable {
+    /// Minimum interval between retries of a session whose last attempt failed.
+    let retryBackoff: TimeInterval
+
+    /// Sessions whose most recent mirror failed and which should be retried.
+    private(set) var pendingSessionIDs: Set<String> = []
+
+    private var nextAttemptID = 0
+    /// Attempt id of the running attempt per session, if any.
+    private var inFlightAttemptIDs: [String: Int] = [:]
+    private var lastFailureDates: [String: Date] = [:]
+    /// Sessions that asked for a mirror while one was already running.
+    private var rescheduleRequests: Set<String> = []
+
+    init(retryBackoff: TimeInterval) {
+        self.retryBackoff = retryBackoff
+    }
+
+    /// Reserve an attempt for `sessionID`.
+    /// - Returns: the attempt id, or `nil` when an attempt is already running.
+    mutating func beginAttempt(sessionID: String) -> Int? {
+        guard inFlightAttemptIDs[sessionID] == nil else { return nil }
+        nextAttemptID += 1
+        inFlightAttemptIDs[sessionID] = nextAttemptID
+        return nextAttemptID
+    }
+
+    /// Record the outcome of an attempt.
+    /// - Returns: `false` when the report comes from a superseded attempt, in
+    ///   which case the caller must not touch any persisted state.
+    mutating func finishAttempt(
+        sessionID: String,
+        attempt: Int,
+        succeeded: Bool,
+        now: Date = Date()
+    ) -> Bool {
+        guard inFlightAttemptIDs[sessionID] == attempt else { return false }
+        inFlightAttemptIDs[sessionID] = nil
+        if succeeded {
+            pendingSessionIDs.remove(sessionID)
+            lastFailureDates[sessionID] = nil
+        } else {
+            pendingSessionIDs.insert(sessionID)
+            lastFailureDates[sessionID] = now
+        }
+        return true
+    }
+
+    /// Pending sessions the retry pass may drive right now: no attempt running,
+    /// and outside the failure backoff window.
+    func retryableSessionIDs(excluding excluded: String? = nil, now: Date = Date()) -> [String] {
+        pendingSessionIDs
+            .filter { sessionID in
+                sessionID != excluded
+                    && inFlightAttemptIDs[sessionID] == nil
+                    && !isBackingOff(sessionID, now: now)
+            }
+            .sorted()
+    }
+
+    /// Note that a mirror was requested while an attempt was already running,
+    /// so the newer content is re-mirrored once that attempt reports.
+    mutating func requestReschedule(sessionID: String) {
+        rescheduleRequests.insert(sessionID)
+    }
+
+    /// Consume any coalesced request recorded during the last attempt.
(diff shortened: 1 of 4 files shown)


yazins-ai:

The self-review on this one is the reason it landed unchanged. The first cut had a genuine hazard that is easy to miss: re-driving every pending mirror from `setNotesFolderPath`, which `startTranscription` calls at every recording start, with no in-flight guard. Against a permanently unreachable mount that grows without bound, and each pass spawns another detached task doing blocking file I/O that can sit on the mount's own timeout — so the retry for an unreachable network folder becomes its own problem. Catching that and bounding it to one attempt per session per five minutes turned an unbounded retry into a predictable one.

The ordering race is the subtler catch, and it is real rather than theoretical: deleting a session drops its in-flight reservation, so a fresh attempt can start while the old one is still blocked, and a late success could then clear a pending flag set by a newer failure — silently dropping the durable retry the whole PR exists to provide. Monotonic attempt ids are the right fix, and making `MirrorAttemptLedger` a plain value type so that race is deterministic in tests rather than timing-dependent is what makes it verifiable.

Two more things worth calling out:

- Finding that `finalizeSession` and the post-batch refresh hand-construct `SessionMetadata` and therefore silently cleared a persisted `mirrorPending`. That would have lost the retry exactly when it mattered — on an app quit.
- Flagging that the same refresh drops `speakerNames` and `customNotesGuidance` in the same way, and deliberately *not* fixing it here to keep the PR to one concern. That is the right call, and it is now filed as a known defect rather than a surprise. Please do open it separately.

Moving the startup scan off the actor was a good catch too — one JSON decode per session directory while actor-isolated would have queued every other repository call behind it at launch.

The folder-split behavior on retry is documented and I agree with the choice: writing to a folder the user has moved away from would be worse.

PR #711, opened 2026-08-30
Fix stale 'Waiting for calendar access' banner when the main window is reopened

thunter009:
## Bug

On a real install with Calendar access granted, the Meetings home shows **"Waiting for calendar access"**. Opening the Settings window repairs it.

## What produces it

`currentAccessState` mapped a missing `CalendarManager` onto a real permission state:

```swift
return container.calendarManager?.accessState ?? .notDetermined
```

A missing manager is an internal condition. It is not a pending permission decision. The two shared one sentinel, so "the manager is not built yet" rendered as "the user has not answered the permission prompt yet".

Opening Settings repairs it because the Calendar tab runs `syncCalendarIntegration()`, which builds the manager on the shared container.

## Fix

**Remove the sentinel conflation.** `currentAccessState` is now `CalendarManager.AccessState?`. Nil means "manager not built" and renders no notice at all. Only a real `.notDetermined` from EventKit shows the waiting copy. The decision is extracted into a pure `HomeTimelineCalendarNotice.resolve(integrationEnabled:accessState:)` and unit-tested. `IdleHomeDashboardView` routes through the same resolver, so the two surfaces cannot drift.

**Repair a missing manager passively.** `AppContainer.ensureCalendarIntegrationReady(enabled:)` builds the manager when it is missing and re-reads system authorization when it is not. It never calls `requestAccess()`, so a background refresh tick cannot raise a TCC dialog out of a windowless app. `updateCalendarIntegration` keeps the single authorization request, tied to the window appearing.

## Retracted: the re-bootstrap diagnosis

An earlier revision claimed the recurring banner came from a **second `AppContainer`**, on the reasoning that `OpenOatsRootApp` is constructed inside the `@main` App's `body` and so could be re-initialised on each evaluation. That was established by reading code. The review asked for runtime confirmation.

I instrumented `bootstrap()` with a call counter and drove the reported repro in-process on the UI-test host: order the main window out, set the activation policy to `.accessory`, then set it back to `.regular` and `makeKeyAndOrderFront` the same window.

```
PROBE bootstrap call #1
PROBE pre-hide  calls=1 windows=1
PROBE hidden    calls=1
PROBE reopened existing window
PROBE final     calls=1
```

`bootstrap()` runs exactly once per process. The window is ordered out and never destroyed, and the reopen re-uses it. The container is the same object throughout. **The diagnosis was wrong, so the memoisation is gone from this PR.**

That retires the updater concern completely. This PR no longer touches `AppContainer.bootstrap()` and no longer changes how many `AppUpdaterController` instances are constructed or started. The whole diff to `AppContainer.swift` is now `ensureCalendarIntegrationReady`.

What I have not pinned down is why the banner returns after a Settings repair, given that the container persists. These changes make the nil state harmless and self-healing whichever path produces it, but I am not claiming to have found that trigger.

## Layout caveat, resolved rather than tested

The previous revision replaced an Optional notice view with a `switch` whose default arm was `EmptyView()`, and I flagged the authorized-case layout as unverified. The notice is the first child of a `VStack(alignment: .leading, spacing: 14)`, so an `EmptyView` that takes a spacing slot would shift the whole timeline. It is an `if let` now, which keeps the original Optional shape. There is no layout change left to verify.

## Follow-ups, deliberately not here

`OpenOatsRootApp` being built inside the `@main` App's `body` is still an odd shape — its `@State` never receives scene-graph storage. The probe shows the hazard is not reachable through the reopen path, so it is latent rather than active. Making `OpenOatsRootApp` the `@main` App, or holding the context in a `let` on the App type, is the clean fix and belongs in its own PR.

`IdleHomeDashboardView` is dead code, never instantiated in Sources, Tests, or UITests. I fixed it rather than deleted it to keep this diff honest. Deleting it is a clean follow-up if you want it gone.

## Verification

- `swift build` — Build complete
- `swift test` — 808 tests, 0 failures
- Rebased on current `main`; previously conflicting, now mergeable

### DIFF
--- a/OpenOats/Sources/OpenOats/App/AppContainer.swift
+++ b/OpenOats/Sources/OpenOats/App/AppContainer.swift
@@ -271,6 +271,25 @@ final class AppContainer {
         }
     }
 
+    /// Passive repair for background refresh loops.
+    ///
+    /// Builds the calendar manager when it is missing and re-reads TCC when it is
+    /// not, but never calls `requestAccess()`. `updateCalendarIntegration` requests
+    /// access whenever it observes `.notDetermined`, so calling it from a poller
+    /// tick could raise a system permission dialog out of a windowless background
+    /// app. Authorization is only ever requested from an explicit user action —
+    /// the Calendar settings toggle.
+    ///
+    /// Never clears the manager: disabling the integration is the toggle's job.
+    func ensureCalendarIntegrationReady(enabled: Bool) {
+        guard enabled else { return }
+        if calendarManager == nil {
+            setCalendarManager(CalendarManager())
+        } else {
+            calendarManager?.refreshFromSystem()
+        }
+    }
+
     /// Recreates the calendar manager so EventKit state is reloaded from scratch.
     /// Use this when permission or visible calendars may have changed in System Settings
     /// and a soft status refresh is not sufficient.
--- a/OpenOats/Sources/OpenOats/Views/ContentView.swift
+++ b/OpenOats/Sources/OpenOats/Views/ContentView.swift
@@ -335,10 +335,12 @@ struct ContentView: View {
             overlayManager.defaults = container.defaults
             miniBarManager.defaults = container.defaults
 
-            // Setup calendar integration before the first await so the home timeline
-            // never transiently shows "Waiting for calendar access" on users who already
-            // granted permission. updateCalendarIntegration is synchronous and safe to
-            // call here.
+            // Set up calendar integration before the first await. A missing manager no
+            // longer reads as "waiting for access" — HomeTimelineCalendarNotice.resolve
+            // treats it as no notice at all — so this is no longer needed to suppress a
+            // flicker. It stays because it is the one call that may request Calendar
+            // authorization, and doing that here keeps the request tied to the window
+            // appearing rather than to a background refresh tick.
             container.updateCalendarIntegration(enabled: settings.calendarIntegrationEnabled)
 
             await container.seedIfNeeded(coordinator: coordinator)
--- a/OpenOats/Sources/OpenOats/Views/HomeTimelineModels.swift
+++ b/OpenOats/Sources/OpenOats/Views/HomeTimelineModels.swift
@@ -145,4 +145,30 @@ enum HomeTimelineCalendarNotice: Equatable {
             return nil
         }
     }
+
+    /// Empty-state copy for the timeline, derived from the same resolution as the
+    /// notice so the two cannot drift apart. A `nil` notice — access granted, or no
+    /// manager built yet — gets neutral copy, never permission-flavoured copy.
+    static func emptyTimelineCopy(
+        integrationEnabled: Bool,
+        accessState: CalendarManager.AccessState?
+    ) -> (title: String, description: String) {
+        switch resolve(integrationEnabled: integrationEnabled, accessState: accessState) {
+        case .integrationOff:
+            return (
+                "No saved meetings yet",
+                "Recorded meetings will appear here even while Calendar integration is off."
+            )
+        case .accessDenied, .waitingForAccess:
+            return (
+                "No saved meetings yet",
+                "Saved meetings will appear here even before Calendar access is available."
+            )
+        case nil:
+            return (
+                "No meetings yet",
+                "Upcoming calendar meetings and saved history will appear here."
+            )
+        }
+    }
 }
--- a/OpenOats/Sources/OpenOats/Views/HomeTimelineModels.swift
+++ b/OpenOats/Sources/OpenOats/Views/HomeTimelineModels.swift
@@ -121,3 +121,28 @@ enum HomeTimelineGrouping {
         }
     }
 }
+
+/// Which calendar notice, if any, the home timeline should show.
+enum HomeTimelineCalendarNotice: Equatable {
+    case integrationOff
+    case accessDenied
+    case waitingForAccess
+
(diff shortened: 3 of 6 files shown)


yazins-ai:
Thanks @thunter009 — and thank you especially for the revision that retracted the first diagnosis. "I went back at the causal story and part of it was wrong" is a rare and genuinely useful thing to find in a PR body, and the second explanation is much stronger: nothing destroys the manager, so a recurring banner has to mean a different container, not a lost one.

I have moved this to `risk:high` / `state:awaiting-human` rather than merging. One specific reason, and it is not the calendar logic:

**The fix changes updater instantiation.** Memoising `AppContainer.bootstrap()` also, in your words, "stops a second `AppUpdaterController` being constructed and started." Repo policy treats any change to updater behavior as maintainer-approved by default, and that applies whether the change makes the updater behave better or worse. Going from two started Sparkle controllers to one is very likely a fix — but Sparkle drives automatic update checks and installation, and an autonomous agent should not be the one to decide that unsupervised.

Two other things a maintainer should weigh:

- The re-bootstrap diagnosis is established by code reading, and you say so plainly. Because the memoisation changes container identity for every scene in the app, its blast radius is much wider than the banner it fixes. A runtime confirmation — even a counter in `bootstrap()` logged on a real launch, which you already suggested — would turn this from a strong inference into a verified one.
- Your own caveat that the authorized-case layout is unverified after swapping the Optional notice for `EmptyView()`.

I would also note your closing suggestion is the better long-term shape: making `OpenOatsRootApp` the `@main` App, or holding the context in a `let` on the App type, removes the hazard at the source instead of memoising around it. Worth @yazinsai deciding between the two here rather than taking the memoisation now and the real fix later.

Everything else in this PR — the sentinel removal, the pure `HomeTimelineCalendarNotice.resolve`, and catching that the earlier revision could raise a TCC dialog from a background tick — is good work and I would happily have merged it on its own.

thunter009:
All three points are addressed, and the first one is addressed by deletion rather than by argument.

**The updater change is gone.** You were right to stop this on that alone. I did the runtime confirmation you asked for, and it disproved the diagnosis that the memoisation rested on.

I put a call counter in `bootstrap()` and drove the reported repro in-process: order the main window out, set the activation policy to `.accessory`, then set it back to `.regular` and `makeKeyAndOrderFront` the same window.

```
PROBE bootstrap call #1
PROBE pre-hide  calls=1 windows=1
PROBE hidden    calls=1
PROBE reopened existing window
PROBE final     calls=1
```

One call per process. The window is ordered out, never destroyed, and the reopen re-uses it, so the container is the same object the whole way through. There is no second `AppContainer` on this path and there never was.

So the memoisation is removed. This PR no longer touches `AppContainer.bootstrap()`, and the number of `AppUpdaterController` instances constructed and started is exactly what it is on `main`. The remaining diff to `AppContainer.swift` is `ensureCalendarIntegrationReady` and nothing else. The blast-radius concern — container identity changing for every scene in the app — goes with it.

Worth saying plainly: my second diagnosis was wrong in the same way the first one was. Both were causal stories built from reading code, and both sounded stronger than the evidence supported. The instrumentation took twenty minutes and would have caught it at either stage.

**The layout caveat is resolved rather than tested.** The `switch` with an `EmptyView()` default is now an `if let`, which preserves the original Optional view shape. Since the notice is the first child of a `VStack(spacing: 14)`, that removes the question of whether `EmptyView` contributes a spacing slot instead of answering it.

**On the long-term shape:** `OpenOatsRootApp()` inside the `@main` App's `body` is still worth fixing, and I agree the `@main`-or-`let` choice is @yazinsai's. The probe changes its priority though — the hazard is latent, not active, so it is a cleanup PR rather than something this bug fix needs to carry.

Also rebased on current `main`; it was conflicting and is not now. `swift test` — 808 tests, 0 failures. I rewrote the PR body: it had two superseded root-cause sections stacked on each other, and the second one was the one I have now retracted.

yazins-ai:
Re-triaged under the updated repo policy — this is now `risk:medium` / `state:ready-to-merge`.

The only thing that held this at `risk:high` was the updater instantiation change, and you removed it. I re-read the diff to confirm rather than taking the summary: the `AppContainer.swift` hunk is `ensureCalendarIntegrationReady` and nothing else, `bootstrap()` is untouched, and the number of `AppUpdaterController` instances is exactly what it is on `main`. The policy clause I invoked no longer has anything to attach to.

Two things independently pushed the same way:

- Your runtime probe. I asked for it as a way to verify the memoisation was safe; it did something more useful by disproving the premise. One `bootstrap` call per process, window ordered out and re-used rather than destroyed. Retracting your own diagnosis twice, with instrumentation instead of argument, is the reason this is merging on the first policy pass rather than the third.
- A policy change on the maintainer's side: file count across subsystems is no longer a `risk:high` trigger on its own, and fixing a permission flow so that it does what it already claims is explicitly carved out. This PR is the second of those.

`validate-swift`, `package-smoke`, and `ui-smoke` are all green and the branch is mergeable with no conflicts. Queued for squash merge.

The `OpenOatsRootApp()`-inside-`body` cleanup is worth doing on its own terms — your read that the probe demotes it from bug fix to latent cleanup matches mine.

yazins-ai:
Merged. Thanks @thunter009 — the instrumentation that disproved your own memoisation diagnosis is what got this to land on the first policy pass, and removing the updater change rather than defending it is why it could be merged without maintainer sign-off. The `ensureCalendarIntegrationReady` split — passive repair on the poller tick, `requestAccess()` only from the settings toggle — is the right shape for that hazard.

PR #721, opened 2026-09-14
Support macOS Sonoma 14.2+

mnm-matin:
Enable the official app to run on Apple Silicon Macs with macOS 14.2+ (Sonoma), while retaining Qwen3 ASR on macOS 15+.

- Align SwiftPM, app and UI-test deployment targets; filter unavailable Qwen3 choices and resolve saved live/batch selections to Parakeet v2 on Sonoma.
- Fix device-bound audio taps: select stream 0 and reject an unknown tap ID even after a successful creation status. This fixes the reproduced `560947818` / “Stream out of range for tap” failure.
- Preserve capture sample rates through delayed VAD/ASR processing, with regression coverage for duration, pitch and format changes.
- Derive Sparkle/Homebrew minimum OS from each released bundle and add deployment checks. Leave the current cask's requirement unchanged until a compatible binary ships.

Includes tests and concise build/device-verification documentation. Build tooling remains Xcode 26 / Swift 6.2; end users on Sonoma do not need Xcode.

Validation on clean Xcode 26.6 / Swift 6.3.3 runners, without dependency patches:

- [Debug/release builds and 828 Swift tests](https://github.com/mnm-matin/OpenOats/actions/runs/34888990577): passed, zero failures.
- [Package smoke test](https://github.com/mnm-matin/OpenOats/actions/runs/34888490161): passed, including packaged minimum-OS checks and four release-metadata tests.
- [10 UI smoke tests](https://github.com/mnm-matin/OpenOats/actions/runs/34888490118): passed.

Package/UI runs use `41da7d4`; the only subsequent change at `7ac06a9` adjusts a unit-test assertion for normal converter priming. Production code is identical.

A local Sonoma package build and production-converter probe also passed. The underlying audio fixes passed a real macOS 14.6.1 microphone-muted playback/transcription test using built-in speakers; this does not claim all-peripheral or long-meeting coverage.


yazins-ai:
Thanks @mnm-matin, this is a careful and well-validated PR. The Sonoma testing on real hardware and the stream-0 / unknown-tap-ID fix for `560947818` are especially useful.

Deception check: clean. The workflow and script changes only read `LSMinimumSystemVersion` from the built bundle and thread it into the appcast and cask. Nothing touches secrets, signing, or permissions.

Duplicate / existing-work check: not already on `main`. `main` targets macOS 15 in `Package.swift`, `Info.plist`, the appcast (`minimumSystemVersion` hard-coded to 15.0), and the cask.

**Labels:** `kind:feature`, `risk:high`, `release:minor`, `state:awaiting-human`. This changes the supported OS floor and edits `release-dmg.yml`, the Sparkle appcast `minimumSystemVersion`, and the Homebrew cask script. Under the repo policy, distribution and release-workflow changes need maintainer sign-off.

**Decision needed:** should OpenOats lower its minimum supported macOS from 15.0 to 14.2 and ship the matching release-pipeline changes?

**Options**
1. **Merge as-is.** Sonoma users get the app and auto-updates. The cost is a permanent 14.2 support floor: future APIs that need macOS 15 must be gated, and Sonoma-only bugs become ours to support.
2. **Split.** Land the system-audio tap fix (stream 0, rejecting an unknown tap ID) and the sample-rate preservation separately at `risk:medium`, and keep the deployment-target and release-pipeline change pending. The bug fixes reach macOS 15 users now, and the OS-floor decision stays open.
3. **Decline the OS-floor change** and keep macOS 15 as the floor. This is the simplest to support, but Sonoma users stay excluded.

**Recommendation: option 1.** The PR has already done the hard work: Qwen3 is gated to macOS 15+, saved selections fall back to Parakeet, packaged minimum-OS checks run in `package-smoke`, and the cask requirement only moves once a compatible binary ships. All three required checks are green. If you would rather not commit to 14.2 yet, option 2 gets the audio fixes out without that decision.

**To approve option 1:** replace `state:awaiting-human` with `state:ready-to-merge`, keeping `risk:high`. For option 2, replace `state:awaiting-human` with `state:planned` and I'll open the split bug-fix PR, crediting @mnm-matin.

yazins-ai:
Merged. Thank you @mnm-matin! Sonoma support is a big reach expansion, and you did it carefully: Qwen3 is gated behind macOS 15 with a Parakeet fallback for saved selections, the Sparkle and Homebrew minimum OS is now read from the built bundle, and `package-smoke` checks that the packaged binary's minimum OS matches. The tap stream-0 / unknown-tap-ID fix for `560947818` should also help macOS 15 users who have seen "Stream out of range for tap". This will ship in the next release.

